PR #1, opened 2017-08-21
Add getPackages and runWorkspaceTasks method to the public API

d4rkr00t:


src/utils/runWorkspaceTasks.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+import type Task from '../Project';
+import Project from '../Project';
+
+export default async function runWorkspaceTasks(task: Task) {
```
d4rkr00t: I'm not sure whether this is a right approach, but if we don't want to expose Project's API it looks fine to me. Also we can keep Project.runWorkspaceTasks api unchanged.

src/utils/runWorkspaceTasks.js
```diff
@@ -0,0 +1,10 @@
...
+  let cwd = process.cwd();
+  let project = await Project.init(cwd);
+  let workspaces = await project.getWorkspaces();
+  Project.runWorkspaceTasks(workspaces, task);
```
thejameskyle: We should `await` this too so that we can rely on the promise it returns.

src/utils/runWorkspaceTasks.js
thejameskyle: Yeah, I like this api

src/utils/runWorkspaceTasks.js
d4rkr00t: oh, thanks :)

src/utils/runWorkspaceTasks.js
d4rkr00t: Also I thought maybe it makes sense to be able to pass CWD as an argument?

d4rkr00t:
@thejameskyle done :)

src/utils/runWorkspaceTasks.js
thejameskyle: Ah yeah, that seems reasonable

thejameskyle: Maybe as an option? 

```js
export default async function runWorkspaceTasks(tasks: Task, opts: { cwd?: string } = {}) {
  // ...
}
```

PR #3, opened 2017-08-27
Lbatch/add update package versions

lukebatchelor:
Adds the updatePackageVersions util function for bumping/releasing

### DIFF
--- a/src/utils/updatePackageVersions.js
+++ b/src/utils/updatePackageVersions.js
@@ -5,20 +5,20 @@ import runWorkspaceTasks from './runWorkspaceTasks';
 
 export default async function updatePackageVersions(versions: {[name: string]: string}, opts: { cwd?: string } = {}) {
   const cwd = opts.cwd || process.cwd();
+  const project = await Project.init(cwd);
+  const workspaces = await project.getWorkspaces();
 
-  const updateWorkspacePackageVersion = (workspace: Workspace) => {
+  for (let workspace of workspaces) {
     const pkg = workspace.pkg;
     const promises = [];
 
-    for(let depName of Object.keys(versions)) {
+    for (let depName of Object.keys(versions)) {
       const depType = pkg.getDependencyType(depName);
       if (!depType) continue;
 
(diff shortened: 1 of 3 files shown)


thejameskyle:
You should export this from https://github.com/pyarn/pyarn/blob/master/src/index.js

src/utils/updatePackageVersions.js
```diff
@@ -0,0 +1,24 @@
...
+      const depType = pkg.getDependencyType(depName);
+      if (!depType) continue;
+
+      promises.push(pkg.updateDependencyVersionRange(depName, depType, versions[depName]));
+    }
+
+    return Promise.all(promises);
+  }
+
+  await runWorkspaceTasks(updateWorkspacePackageVersion, opts);
```
thejameskyle: You don't need `runWorkspaceTasks` for this, it's meant only for scripting.

lukebatchelor: Should this take in a Workspace then?

thejameskyle: Nah, we should not expose internal objects like `Workspace` or `Package`. This is all you need to do:

```
let cwd = process.cwd();

for (let workspace of workspaces) {
  // ...
}
```

src/utils/updatePackageVersions.js
```diff
@@ -0,0 +1,24 @@
...
+
+  for (let workspace of workspaces) {
+    const pkg = workspace.pkg;
+    const promises = [];
+
+    for (let depName of Object.keys(versions)) {
+      const depType = pkg.getDependencyType(depName);
+      if (!depType) continue;
+
+      promises.push(pkg.updateDependencyVersionRange(depName, depType, versions[depName]));
```
thejameskyle: This should probably be:

```js
'^' + versions[depName]
```

That way the version ranges that we're publishing can match a wider range

thejameskyle:
Fix that one thing and feel free to merge

src/utils/updatePackageVersions.js
lukebatchelor: Oh, in testing, I was adding the hat in myself.

PR #4, opened 2017-08-27
Pyarn publish command

lukebatchelor:
Probably dont need to merge just yet, just wanted to get this up to discuss a couple of things.


src/utils/locks.js
```diff
@@ -0,0 +1,35 @@
...
+      locks.push(pkg);
+    });
+    promises.push(promise);
+  }
+
+  try {
+    await Promise.all(promises);
+  } catch (err) {
+    await unlock(locks);
+    throw new Error('Found a lock tag', err);
```
lukebatchelor: I added the `, err` here because I wanted to show the more detailed error message and wasnt sure whether to move it here, or have it closer to the source of the problem.

src/utils/npm.js
```diff
@@ -1,18 +1,50 @@
...
+
+
+export async function removeTag(pkgName: string, tag: string) {
+  return npmRequestLimit(async () => {
+    logger.info(`npm dist-tag rm ${pkgName} ${tag}`);
+
+    try {
+      return await processes.spawn('npm', ['dist-tag', 'rm', pkgName, tag]);
+    } catch (error) {
+      // The dist tag not existing is unexpected, but shouldn't prevent execution
```
lukebatchelor: Thoughts on this?
It's bad that we are using an error message as "api", but better than halting because you expected a tag?

src/utils/npm.js
```diff
@@ -1,18 +1,50 @@
...
     return JSON.parse(result.stdout);
   });
 }
+
+export function addTag(pkgName: string, pkgVersion: string, tag: string) {
+  return npmRequestLimit(async () => {
+    const pkgStr = `${pkgName}@${pkgVersion}`;
+    logger.info(`npm dist-tag ${pkgStr} ${tag}`);
+
+    const result = await processes.spawn('npm', ['dist-tag', 'add', pkgStr, tag]);
```
lukebatchelor: There's an interesting case here, not sure if we need to worry about it or not.
If a tag already exists but at a different version of the package, this command wont throw any errors.

It's *probably* not a huge issue, as you'd have to be behind master for this to ever occur (or someone else has manually published and that change wasnt in master).

The alternative would be to do an `npm dist-tag ls` but that just opens up more chance for race conditions, so probably not important.

src/commands/publish-unlock.js
```diff
@@ -0,0 +1,15 @@
...
+export default async function publish() {
+  let cwd = process.cwd();
+  let project = await Project.init(cwd);
+  let workspaces = await project.getWorkspaces();
+
+  let packages = workspaces.map(workspace => workspace.pkg);
+
+  await locks.unlock(packages);
+
+  // ...
```
lukebatchelor: Was there anything else required here?

src/utils/getPackages.js
```diff
@@ -5,5 +5,9 @@ export default async function getPackages(opts: { cwd?: string } = {}) {
   let cwd = opts.cwd || process.cwd();
   let project = await Project.init(cwd);
   let workspaces = await project.getWorkspaces();
-  return workspaces.map(workspace => workspace.pkg.dir);
+  return workspaces.map(workspace => ({
```
lukebatchelor: Updated as discussed. 
Cant see any usages of this, and @d4rkr00t  hasn't used it anywhere, so should be a safe change

src/utils/getUnpublishedPackages.js
```diff
@@ -0,0 +1,24 @@
...
+export default async function getUnpublishedPackages(opts: { cwd?: string } = {}) {
+  const cwd = opts.cwd || process.cwd();
+  const project = await Project.init(cwd);
+  const workspaces = await project.getWorkspaces();
+  const unpublishedPackages = [];
+
+  for (let workspace of workspaces) {
+    const packageInfo = await npm.info(workspace.pkg.config.name);
+
+    if (packageInfo.version !== workspace.pkg.config.version) {
```
lukebatchelor: Should this be a `semver.gt()` instead of `!==` do you reckon?

src/utils/locks.js
lukebatchelor: Oh, silly. I Cant do that.
Do you have an opinion for where this should be handled?
We *need* to catch it higher, but it makes sense to keep the error message closer to where the error occured.
Should this just rethrow the same error after unlocking to you think?

src/utils/npm.js
lukebatchelor: As discussed, will add an npm-info check here as an extra precaution.

src/commands/publish.js
```diff
@@ -1,6 +1,37 @@
 // @flow
 import type {Args, Opts} from '../types';
+import {PError} from '../utils/errors';
+import * as logger from '../utils/logger';
+import * as locks from '../utils/locks';
+import getUnpublishedPackages from '../utils/getUnpublishedPackages';
+import Project from '../Project';
+import * as npm from '../utils/npm';
+import * as yarn from '../utils/yarn';
```
lukebatchelor: - [ ] Remove

src/commands/publish.js
```diff
@@ -1,6 +1,37 @@
...
+  const cwd = process.cwd();
+  const project = await Project.init(cwd);
+  const workspaces = await project.getWorkspaces();
+  const packages = workspaces.map(workspace => workspace.pkg);
+
+  try {
+    await locks.lock(packages);
+    const unpublishedPackages = await getUnpublishedPackages(opts);
+    const isUnpublished = workspace => unpublishedPackages.some(pkg => workspace.pkg.config.name === pkg.name);
+    const unpublishedWorkspaces = workspaces.filter(isUnpublished);
```
lukebatchelor: Wasn't super happy with this little dance here
getWorkspaces -> getPackages -> lockPackages -> getUnpublishedPackages -> use those to filter original workspaces.

src/commands/publish.js
```diff
@@ -1,6 +1,37 @@
...
+
+    await Project.runWorkspaceTasks(unpublishedWorkspaces, async workspace => {
+      const {name, version} = workspace.pkg.config;
+      logger.info(`Publishing ${name} at ${version}`);
+
+      await npm.publish(workspace.pkg.config.name, { cwd: workspace.pkg.dir });
```
lukebatchelor: - [ ] Change to `name`

src/commands/publish.js
```diff
@@ -1,6 +1,37 @@
...
+    });
+
+    await locks.unlock(packages);
 
+  } catch (err) {
+    logger.error(err.message);
```
lukebatchelor: We *could* add an
```
await locks.unlock(packages)
```

here, but it would add a lot more noise when you hit the code path where you roll back already

attemptToGetLocks -> somePass, some  fail -> clean up all locks -> rethrow error -> gets caught here -> cleans up again (which is super loud because npm puts three lines of warning on the stderr per package)

Fine for now, but would like to make cleaner

src/utils/getPackages.js
```diff
@@ -5,5 +5,10 @@ export default async function getPackages(opts: { cwd?: string } = {}) {
   let cwd = opts.cwd || process.cwd();
   let project = await Project.init(cwd);
   let workspaces = await project.getWorkspaces();
-  return workspaces.map(workspace => workspace.pkg.dir);
+
+  return workspaces.map(workspace => ({
```
lukebatchelor: Changed API as we discussed, look right?

src/utils/locks.js
```diff
@@ -0,0 +1,49 @@
...
+
+export async function lock(packages: Array<Package>) {
+  let locks = [];
+  let promises = [];
+
+  logger.info('Attempting to get locks for all packages');
+
+  for (let pkg of packages) {
+    let {name, version} = pkg.config;
+    let promise = npm.info(name).then(pkgInfo => {
```
lukebatchelor: This is the reason we cant *easily* cache the npm.info command, we sometimes want data that could have changed without a version bump (like dist tags).

We could provide two methods though?

src/utils/npm.js
lukebatchelor: Ended up adding this check to locks.lock()

src/utils/promises.js
```diff
@@ -0,0 +1,39 @@
+/**
```
lukebatchelor: Happy to remove all this if you like.

src/commands/publish.js
thejameskyle: The way that I was thinking of handling rollbacks is to do stuff in caught phases:

```js
try {
  await doSomeWork(); // phase 1
  try {
    await doSomeMoreWork(); // phase 2
    try {
      await doFinalWork(); // phase 3
    } catch (err) {
      await undoFinalwork(); // rollback phase 3
      throw err; // rethrow to bubble
    }
  } catch (err) {
    await undoSomeMoreWork(); // rollback phase 2
    throw err; // rethrow to bubble
  }
} catch (err) {
  await undoSomeWork(); // rollback phase 1
  throw err; // rethrow to bubble
}
```

Just as long as we abstract each "phase" into its own function we will avoid this hell, and we can rely on each step rolling itself back and throwing.

I'm hopefully this will work, but I might be wrong– Particularly if we need to rollback in an order other than the reverse order (which I don't think we need to).

thejameskyle: You could also log in these places as the re-thrown errors bubble up to report what is happening

src/utils/locks.js
thejameskyle: I think you can ignore caching for now

src/utils/npm.js
thejameskyle: Should we be adding this logic in this function?

You could always have:

```js
npm.removeTag(); // throws a nice error when there's no dist-tag
npm.removeTagSafe(); // does not throw an error, logs
```

src/utils/npm.js
```diff
@@ -1,18 +1,58 @@
...
+export async function removeTag(pkgName: string, tag: string) {
+  return npmRequestLimit(async () => {
+    logger.info(`npm dist-tag rm ${pkgName} ${tag}`);
+
+    try {
+      return await processes.spawn('npm', ['dist-tag', 'rm', pkgName, tag]);
+    } catch (error) {
+      // The dist tag not existing is unexpected, but shouldn't prevent execution
+      if (error.code === 1 && error.stderr.includes('is not a dist-tag on')) {
+        logger.info(`No dist tag "${tag}" found for package ${pkgName}`);
```
thejameskyle: This should probably be a warning instead of info

src/utils/promises.js
```diff
@@ -0,0 +1,39 @@
...
+      error = err;
+    }
+  }
+
+  if (error) {
+    throw error;
+  } else {
+    return results;
+  }
+}
```
thejameskyle: newline

src/commands/install.js
```diff
@@ -95,6 +95,8 @@ export default async function install(args: Args, opts: Opts) {
 
   logger.log('[2/2] Linking workspace dependencies...');
 
+  await fs.mkdirp(project.pkg.nodeModulesBin);
```
thejameskyle: Should we create this directory or should we make a `fs.readdirSafe()` that returns an empty array when the directory does not exist?

lukebatchelor: I figured an empty directory wouldnt be a dangerous thing to have, happy to update to readdirSafe though, since we already wrap all the fs functions as it is

src/utils/npm.js
lukebatchelor: Sure, we can. Which would you want to call in this case though?

src/utils/npm.js
lukebatchelor: Sure. It's already super noisy with the warnings above though.
But semantically, this probably makes more sense.

src/utils/promises.js
lukebatchelor: Dammit. I'm gonna add an eslint file to the root and just gitignore it. I need my auto-fixing

src/commands/install.js
lukebatchelor: I added the function but didnt end up actually needing it. I think some of the work you did with bin linking recently must have fixed it?

PR #5, opened 2017-08-31
Exposes getDependencyGraph function to consumers

lukebatchelor:
As discussed, removes all internal data structures and returns in the form:

```
const pyarn = require('pyarn');
const depGraph = pyarn.getDependencyGraph();// uses cwd by default, can be overridden

console.log(depGraph.get('package-a')) // ['package-b', 'package-c'];
```

Opinion on having this be a map vs an object?


src/utils/getDependencyGraph.js
```diff
@@ -0,0 +1,24 @@
+// @flow
+import Project from '../Project';
+import * as yarn from '../utils/yarn';
+
+export default async function getDependencyGraph(opts: { cwd?: string } = {}) {
+  const cwd = process.cwd();
```
thejameskyle: opts.cwd?

PR #6, opened 2017-08-31
Prevents running scripts in packages if they don't exist

lukebatchelor:
I ran into a bunch of issues with preinstall and postinstall scripts not being present.

We are getting the same issue in CI too. For Trey, it seems to happen in node 8.4.0 and not 8.1.4. 
It seems to happen to me in both versions (and CI is running 8.1.4).

In either case this seemed logical?


thejameskyle:
Ah yeah, I wrapped `run` with another function previously to do this check

PR #7, opened 2017-08-31
Introduces filtering by name and filepath

lukebatchelor:
Supported on `pyarn workspaces run` command in this PR, will add to everything else if you're happy with it like this.

Also exposed as a utility that can be imported, thinking about it, that would expose the Package implementation details. So, happy to remove or start separating lib/ and util/ commands so we can have exposed/simplified versions in lib/


lukebatchelor:
```
$ yarn run build && (cd example-2 && ../bin.js workspaces run postinstall --ignore "*/bar" --only "**/baz")
pyarn workspaces v0.7.0
(@lbatchelor/baz) $ yarn run postinstall -s yarn run v0.22.0
(@lbatchelor/baz) $ yarn run postinstall -s $ echo postinstall
(@lbatchelor/baz) $ yarn run postinstall -s postinstall
(@lbatchelor/baz) $ yarn run postinstall -s Done in 0.07s.
```

thejameskyle:
Do we want to warn in case there's zero matched packages?

I don't think it should be an error, but a `logger.warn` makes sense to me

src/utils/filterWorkspaces.js
```diff
@@ -0,0 +1,28 @@
+// @flow
+import multimatch from 'multimatch';
+import path from 'path';
+
+import type {Args, Opts} from '../types';
+import Workspace from '../Workspace';
+
+export default function filterWorkspaces(workspaces: [Workspace], opts: Opts) {
```
thejameskyle: Assuming this type should be `Array<Workspace>`

src/index.js
```diff
@@ -2,6 +2,7 @@
 export {default as run} from './lib';
 export {default as cli} from './cli';
 export {default as getPackages} from './utils/getPackages';
+export {default as filterWorkspaces} from './utils/filterWorkspaces';
```
thejameskyle: We should not expose this function so that it requires a `Workspace` object to be passed

lukebatchelor: No worries, will remove for now.

Do you think it'd be worth splitting the utils into ones that are exposed and ones that arent, and having the ones that are exposed give you the simplified output like we do with getPackages?

src/utils/__tests__/filterWorkspaces.test.js
```diff
@@ -0,0 +1,62 @@
...
+    })
+
+    it('should using both --only and --ignore', () => {
+      const filteredWorkspaces = filterWorkspaces(allWorkspaces, { only: '**/ba*', ignore: 'baz' });
+      expect(filteredWorkspaces.length).toEqual(1);
+      expect(workspacesContains(filteredWorkspaces, ['@scope/bar'])).toEqual(true);
+    })
+  });
+
+  describe('filtering by path', () => {
```
lukebatchelor: I've run into an issue here.

FilterWorkspaces uses the cwd to find the relative paths to each package so that it can filter by path (i.e --onlyFs="packages/foo"). This is fine when you're actually running commands (assuming you're running at the root of the project), but obviously doesnt work in tests.

There are a couple of solutions... We could pass the cwd to filterWorkspaces... That's a bit messy.

Or maybe we could add relativeDir to each package along with dir? That would be a lot simpler imo and might be useful later on.

src/utils/filterWorkspaces.js
```diff
@@ -0,0 +1,39 @@
+// @flow
+import multimatch from 'multimatch';
+import path from 'path';
+
+import Workspace from '../Workspace';
+import * as logger from '../utils/logger';
+
+type FilteringOpts = {
```
lukebatchelor: So, this isnt great, wondering if you know a better way?
Flow yells because a flag can either be a string or a bool and supposedly (???) it can't coerce a bool to a string, so it yells on lines 16-19.

Adding this type here though, means you need to create a separate opts object any time you want to filter, that's pretty crap too.

Could we add explicit types for the options we know will be strings perhaps?

thejameskyle: Technically options could end up inside commands as strings or booleans since there is nothing to validate them... We need a better way to do that.

In the meantime I've just been doing stuff like:

```js
let cwd = typeof opts.cwd === 'string' ? opts.cwd : process.cwd();
```

src/commands/ws/run.js
```diff
@@ -2,15 +2,17 @@
 import type {Args, Opts} from '../../types';
 import Project from '../../Project';
 import * as yarn from '../../utils/yarn';
+import filterWorkspaces from '../../utils/filterWorkspaces';
 
 export default async function run(args: Args, opts: Opts) {
   let cwd = process.cwd();
   let project = await Project.init(cwd);
   let workspaces = await project.getWorkspaces();
+  let filteredWorkspaces = filterWorkspaces(workspaces, opts);
```
thejameskyle: I'd really prefer if we manually specified which options got passed between functions everywhere. I don't like passing unrelated objects around.

In general I think you should treat all "options" or "config" in Flow as ["Exact types"](https://flow.org/en/docs/types/objects/#toc-exact-object-types)

```js
function fn1(opts1: {| foo: string, bar: string |}) {
  fn2(opts1); // Error: `bar` is not allowed in type of `opts2`
}

function fn2(opts2: {| foo: string |}) {
  // ...
}
```

Instead, if you specify the options manually, it will work:

```js
fn2({ foo: opts.foo }); // Works
```

lukebatchelor: Yup, for sure

src/utils/__tests__/filterWorkspaces.test.js
lukebatchelor: Going to move into Project 👍

lukebatchelor:
Superceded by https://github.com/pyarn/pyarn/pull/25

PR #8, opened 2017-09-01
Fix flow types

thejameskyle:
cc @lukebatchelor


src/utils/npm.js
```diff
@@ -22,7 +22,9 @@ export function publish(pkgName: string, opts: { cwd?: string } = {}) {
   return npmRequestLimit(async () => {
     logger.info(`npm publish ${pkgName}`);
 
-    return await processes.spawn('npm', ['publish'], opts);
+    return await processes.spawn('npm', ['publish'], {
+      cwd: opts.cwd,
```
thejameskyle: @lukebatchelor btw I think we should be more explicit when passing options around, passing an entire `opts` object into another could cause subtle bugs

PR #9, opened 2017-09-04
Add tests for config

thejameskyle:
cc @lukebatchelor


src/utils/__fixtures__/package-with-uncertain-indentation-json/package.json
```diff
@@ -0,0 +1,3 @@
+{
+"name": "four-spaces"
```
lukebatchelor: Prrrooobably nothing, but is this supposed to have four spaces?

lukebatchelor: Ah, I see. It will override this one with 2 spaces

lukebatchelor:
Sweet, LGTM, I'll follow a similar structure with the workspaces commands

PR #10, opened 2017-09-05
Add tests for fs util

thejameskyle:
This took forever and the symlink tests aren't amazing, but here are some tests for fs


thejameskyle:
cc @lukebatchelor

PR #11, opened 2017-09-06
Run npm.info in getUnpublishedPackages in parallel

thejameskyle:
This was running serially, this should speed it up quite a bit

@lukebatchelor


lukebatchelor:
Oh sweet. Nice

PR #12, opened 2017-09-06
Make new directory for public functions

thejameskyle:
@lukebatchelor


src/functions/runWorkspaceTasks.js
```diff
@@ -0,0 +1,27 @@
...
+  config: Config,
+};
+
+type Task = (workspace: WorkspaceInfo) => Promise<mixed>;
+
+type Options = {
+  cwd?: string,
+};
+
+export default async function runWorkspaceTasks(task: Task, opts: Options = {}): Promise<void> {
```
thejameskyle: @lukebatchelor Just to note, I was thinking that everything in this directory could explicitly state all the types for input and output so we don't break these unintentionally

src/utils/publish-lock.js
```diff
@@ -1,13 +0,0 @@
-// @flow
```
thejameskyle: This file didn't seem to be used for anything... 🤷‍♀️

src/functions/runWorkspaceTasks.js
lukebatchelor: For sure, sounds good.

src/utils/publish-lock.js
lukebatchelor: .... ... Hmmm.. I think at some point I must have copy + pasted it in VSCode and not noticed it. Its a copy of the file from commands (not sure they /should/ live there either, but thats a different point)

src/functions/runWorkspaceTasks.js
```diff
@@ -0,0 +1,27 @@
+// @flow
+import Project from '../Project';
+import type {Config} from '../types';
+
+type WorkspaceInfo = {
```
lukebatchelor: Fine to leave here for now, but reckon we need a types.js file in /functions as well? Likely types like this will be shared I think.

lukebatchelor:
LGTM, will follow this pattern

src/functions/runWorkspaceTasks.js
thejameskyle: Add it when we need it

PR #15, opened 2017-09-08
Fixes some small issues in publish command

lukebatchelor:
Havent added tests yet, just wanted to get your opinion.

### DIFF
--- a/src/commands/publish.js
+++ b/src/commands/publish.js
@@ -26,14 +26,15 @@ async function getUnpublishedPackages(packages) {
 
     return {
       name: config.name,
-      previousVersion: config.version,
-      version: response.response !== '404' ? response.data.version : 'UNPUBLISHED',
+      localVersion: config.version,
+      isPublished: response.published,
+      publishedVersion: response.pkgInfo.version || '',
     };
   }));
 
   return results.filter(result => {
     // only publish if our version is higher than the one on npm
-    return result.version === 'UNPUBLISHED' || semver.gt(result.previousVersion, result.version);
+    return !result.isPublished || semver.gt(result.localVersion, result.publishedVersion);
   });
 }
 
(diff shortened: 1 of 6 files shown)


src/commands/publish.js
```diff
@@ -8,48 +9,56 @@ import Project from '../Project';
 
 export type PublishOptions = {|
   cwd?: string,
+  access?: string,
 |};
 
 export function toPublishOptions(args: options.Args, flags: options.Flags): PublishOptions {
   return {
     cwd: options.string(flags.cwd, 'cwd'),
+    access: options.string(flags.access, 'access')
```
lukebatchelor: Maybe you'll be able to answer this, but when I first tried to publish I got errors because it was a scoped package so you have to manually add the --access public flag. After it was published though, it didnt need that flag anymore? Is it only required on initial publish?

src/utils/locks.js
```diff
@@ -42,7 +44,9 @@ export async function unlock(packages: Array<Package>) {
   let promises = [];
 
   for (let pkg of packages) {
-    promises.push(npm.removeTag(pkg.config.name, LOCK_DIST_TAG));
+    const promise = npm.removeTag(pkg.config.name, LOCK_DIST_TAG)
```
lukebatchelor: Will revert

src/commands/publish.js
```diff
@@ -8,48 +9,56 @@ import Project from '../Project';
...
+  const results = await Promise.all(packages.map(async pkg => {
+    let config = pkg.config;
+    let response = await npm.infoAllow404(config.name);
 
     return {
-      name: packageInfo.name,
-      version: packageInfo.version,
+      name: config.name,
       previousVersion: config.version,
+      version: response.response !== '404' ? response.data.version : 'UNPUBLISHED',
```
thejameskyle:
```js
{
  // ...
  published: boolean,
  version: string | null
}
```

thejameskyle: `pkgData`

PR #17, opened 2017-09-12
Adds pyarn add functionality

lukebatchelor:
I've just pushed it quickly for you to look at, there might still be some debugging junk that I havent removed.

### DIFF
--- a/src/commands/add.js
+++ b/src/commands/add.js
@@ -3,6 +3,7 @@ import * as logger from '../utils/logger';
 import * as options from '../utils/options';
 import * as yarn from '../utils/yarn';
 import { PError } from '../utils/errors';
+import { symlinkWorkspaces } from '../utils/install';
 import Project from '../Project';
 
 export type AddOptions = {
@@ -60,4 +61,7 @@ export async function add(opts: AddOptions) {
       installedVersion
     );
   }
+
+  // Now we can symlink packages back to the Project package
+  await symlinkWorkspaces({ cwd });
 }
--- a/src/commands/add.js
+++ b/src/commands/add.js
@@ -1,13 +1,63 @@
 // @flow
+import * as logger from '../utils/logger';
 import * as options from '../utils/options';
-import {PError} from '../utils/errors';
+import * as yarn from '../utils/yarn';
+import { PError } from '../utils/errors';
+import Project from '../Project';
 
-export type AddOptions = {};
+export type AddOptions = {
+  pkgs: Array<string>,
+  cwd?: string,
+  dev?: boolean
+};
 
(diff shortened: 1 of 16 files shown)


lukebatchelor:
Ah... I havent actually re-linked bins yet...

lukebatchelor:
Done. Should be good now.
Didnt add tests for utils/install yet though, wont do in this PR if that's okay.

Originally tried to make some lower level functions to be able to link individual packages and workspaces but figured it's actually much easier to just re-link everything, means you'll also get the warnings if something weird happens.

thejameskyle:
Can you update the normalize command with the method rename?

src/utils/yarn.js
```diff
@@ -2,7 +2,17 @@
 import type Package from '../Package';
 import * as processes from './processes';
 
-export async function run(pkg: Package, script: string, args: Array<string> = []) {
+export async function add(pkg: Package, args: Array<string> = []) {
```
thejameskyle: I'd prefer if this method accepted specific options instead of `args` like:

```js
add(pkg, {
  name: 'dep-1',
  version?: '^1.2.3',
  type: 'devDependencies'
});
```

- The `yarn.add` util should be converting JS options into Yarn CLI commands
- The `pyarn add` command should be converting pyarn CLI command into JS options

src/commands/add.js
```diff
@@ -1,13 +1,67 @@
...
+
+    await workspaceToRunIn.pkg.setDependencyVersionRange(
+      pkg,
+      depType,
+      installedVersion
+    );
+  }
+
+  // Now we can symlink packages back to the Project package
+  await symlinkWorkspaces({ cwd });
```
thejameskyle: We probably shouldn't be re-symlinking everything. That's really inefficient and unnecessary. Also it will break locally linked packages (`pyarn link`)

src/utils/__tests__/install.test.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+import * as install from '../install';
+
+describe('util/install', () => {
+  it('should', () => {
+    expect(true).toEqual(true);
```
thejameskyle: perfect

src/commands/add.js
lukebatchelor: Ahhh... Damn.....

Yes

lukebatchelor:
Sweet, will make the changes above.

A thought though, I've just realised, we also need to check that the package you are trying to add isnt one in your project. If it is, it should not run `yarn add` at the project root.

I'll add that too.

lukebatchelor:
Going to decline and rework onto master based on the upcoming remove PR

PR #18, opened 2017-09-14
Exposes getDependentsGraph public function

lukebatchelor:
Not sure why I forgot, but here it is.

Would you also mind doing a release after this one?

### DIFF
--- a/src/__tests__/Project.test.js
+++ b/src/__tests__/Project.test.js
@@ -83,7 +83,6 @@ describe('Project', () => {
       const workspaces = await project.getWorkspaces();
       const { valid, graph } = await project.getDependentsGraph(workspaces);
       const expectedDependents = {
-        'fixture-project-nested-workspaces': [],
         bar: ['foo', 'baz'],
         foo: [],
         baz: []
@@ -143,7 +142,6 @@ describe('Project', () => {
       const workspaces = await project.getWorkspaces();
       const { valid, graph } = await project.getDependentsGraph(workspaces);
       const expectedDependents = {
-        'nested-workspaces-transitive-dependents': [],
         'pkg-a': ['workspace-a', 'pkg-b'],
         'workspace-a': [],
         'pkg-b': ['pkg-c'],
--- a/src/__tests__/Project.test.js
+++ b/src/__tests__/Project.test.js
@@ -6,6 +6,16 @@ import Workspace from '../Workspace';
 
 import { getFixturePath } from 'jest-fixtures';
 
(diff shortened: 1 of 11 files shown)


src/__tests__/index.test.js
```diff
@@ -0,0 +1,22 @@
+import * as functions from '../index';
```
lukebatchelor: Added this, thinking it might be useful.

Was thinking we could also assert on the function signatures/return types, but I guess Flow does that for you?

src/index.js
```diff
@@ -1,5 +1,8 @@
 // @flow
-export {default as getWorkspaces} from './functions/getWorkspaces';
-export {default as getDependencyGraph} from './functions/getDependencyGraph';
-export {default as runWorkspaceTasks} from './functions/runWorkspaceTasks';
-export {default as updatePackageVersions} from './functions/updatePackageVersions';
+export { default as getWorkspaces } from './functions/getWorkspaces';
+export { default as getDependencyGraph } from './functions/getDependencyGraph';
+export { default as getDependentsGraph } from './functions/getDependentsGraph';
+export { default as runWorkspaceTasks } from './functions/runWorkspaceTasks';
+export {
```
lukebatchelor: Prettier. :fistshake:

src/__tests__/Project.test.js
```diff
@@ -52,38 +62,99 @@ describe('Project', () => {
 
     it('should be able to getDependencyGraph', async () => {
       const workspaces = await project.getWorkspaces();
-      const deps = await project.getDependencyGraph(workspaces);
-      const { valid, graph } = deps;
+      const { valid, graph } = await project.getDependencyGraph(workspaces);
+      const expectedDependencies = {
```
lukebatchelor: Refactored all these tests to be more explicit as I managed to change an implementation at one point, without breaking a test.

src/__tests__/index.test.js
thejameskyle: I'm not sure how you would even assert the function signatures other than rewriting the unit tests for them. I don't think we need these tests at all

src/index.js
thejameskyle: lol, embrace syntax apathy

src/__tests__/index.test.js
lukebatchelor: No worries, will remove.

lukebatchelor: Will merge this one now and remove in the next PR which I'm pushing now

PR #19, opened 2017-09-18
Re-exposes the run function

lukebatchelor:
Required so that we can run publish in mk-2.

Also fixes an embarrassing regression in publish. No idea how I missed that...

Also removes the extra test introduced in #18


src/run.js
```diff
@@ -1,53 +1,63 @@
...
+  upgradeInteractive
+} from './commands/upgradeInteractive';
+import { toVersionOptions, version } from './commands/version';
+import { toVersionsOptions, versions } from './commands/versions';
+import { toWhyOptions, why } from './commands/why';
+import { toWorkspaceOptions, workspace } from './commands/workspace';
+import { toWorkspacesOptions, workspaces } from './commands/workspaces';
 
 export default async function pyarn(args: options.Args, flags: options.Flags) {
+  if (!Array.isArray(args)) {
```
lukebatchelor: Decided this was important enough to add.

I got an interesting error where I was running

```
pyarn.run('publish', { access: 'public' };
```

And was getting an error about `run` not being implemented. Turns out it was because the command was getting set to `p`, which is an alias for project, and it was getting treated as `pyarn project ublish`, which becomes `pyarn project run ublish`.

Happy to change the message though

src/utils/locks.js
```diff
@@ -14,18 +14,20 @@ export async function lock(packages: Array<Package>) {
   logger.info('Attempting to get locks for all packages');
 
   for (let pkg of packages) {
-    let {name, version} = pkg.config;
+    let { name, version } = pkg.config;
     let promise = npm.infoAllow404(name).then(response => {
-      if (response.response !== '404') {
-        const pkgInfo = response.data || {};
+      if (response.published) {
```
lukebatchelor: Will need tests for the rest of these util functions **soon** and this wouldnt have happened :(

Would adding an explicit return type to infoAllow404 have helped as well?

thejameskyle:
The reason I removed it was because I was thinking about exposing the command functions directly:

```js
// @flow
import * as pyarn from 'pyarn';

export async function main() {
  let cwd = process.cwd();
  await pyarn.install({ cwd });
  await pyarn.run({ script: 'test', cwd });
  await pyarn.publish({ cwd });
}
```

thejameskyle:
Superseded by #20

PR #25, opened 2017-09-19
Adds Project.filterWorkspaces

lukebatchelor:
Adds filtering and tests.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,10 +1,15 @@
 'use strict';
 
+require('array-includes/polyfill');
 var ver = process.versions.node;
 var majorVer = parseInt(ver.split('.')[0], 10);
 
 if (majorVer < 4) {
-  console.error('Node version ' + ver + ' is not supported in pyarn, please use Node.js 4.0 or higher.');
+  console.error(
+    'Node version ' +
+      ver +
+      ' is not supported in pyarn, please use Node.js 4.0 or higher.'
+  );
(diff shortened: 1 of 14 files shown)


index.js
```diff
@@ -1,10 +1,15 @@
 'use strict';
 
+require('array-includes/polyfill');
```
thejameskyle: Maybe just use this as a library so you don't have to setup jest to polyfill this separately:

```js
import includes from 'array-includes';
```

src/cli.js
```diff
@@ -83,11 +83,11 @@ const commandMap = {
   WORKSPACE_REMOVE: { remove: true, rm: true },
   WORKSPACE_RUN: { run: true },
   WORKSPACE_UPGRADE: { upgrade: true },
-  WORKSPACES: { workspace: true, w: true },
+  WORKSPACES: { workspaces: true, ws: true },
```
lukebatchelor: The only change to this file was actually here, updated to say `workspaces` and `ws`, the rest is just prettier.
(do you have a custom prettier config btw?)

lukebatchelor:
Have added filtering to `workspaces run` now

```
$ (yarn run build:modern && cd example-2 && ../bin.js ws run preinstall --only "**/ba*" --ignore "**/baz")

pyarn v0.11.1
preinstall
Done in 0.62s.
```

PR #26, opened 2017-09-21
Add pyarn remove commands

thejameskyle:
```
pyarn remove dep
pyarn project remove dep
pyarn workspace remove dep
pyarn workspaces remove dep
```

### DIFF
--- a/src/Package.js
+++ b/src/Package.js
@@ -66,33 +66,61 @@ export default class Package {
     return allDependencies;
   }
 
-  async updateDependencyVersionRange(
+  async setDependencyVersionRange(
     depName: string,
     depType: string,
-    versionRange: string
+    versionRange: string | null
   ) {
-    let prevVersionRange =
-      this.config[depType] && this.config[depType][depName];
+    let prevDeps = this.config[depType];
+    let prevVersionRange = (prevDeps && prevDeps[depName]) || null;
     if (prevVersionRange === versionRange) return;
+
+    await this.updateDependencies(depType, {
+      ...prevDeps,
+      [depName]: versionRange
+    });
+
+    let pkgName = this.config.name;
+
+    if (versionRange === null) {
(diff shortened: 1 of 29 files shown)


src/commands/workspace/__tests__/remove.test.js
```diff
@@ -1,4 +1,46 @@
...
+    expect(yarn.remove).toHaveBeenCalledTimes(0);
+    expect(
+      await pathExists(path.join(workspaceDir, 'node_modules', 'foo-dep'))
+    ).toBe(false);
+    expect(
+      await pathExists(path.join(tempDir, 'node_modules', 'foo-dep'))
+    ).toBe(true);
+  });
+
+  test('removing a workspace dependency that doesnt exist in that package', async () => {
```
lukebatchelor: Should we have a test for running from a workspace that isnt the Project as well?

src/commands/__tests__/remove.test.js
```diff
@@ -22,7 +22,7 @@ describe('pyarn remove', () => {
     expect(yarn.remove).toHaveBeenCalledWith(['project-only-dep'], tempDir);
   });
 
-  test('removing a workspace dependency', async () => {
+  test.only('removing a workspace dependency', async () => {
```
lukebatchelor: .only

src/Package.js
```diff
@@ -66,33 +66,61 @@ export default class Package {
...
+          pkgName,
+          depName,
+          versionRange,
+          prevVersionRange
+        )
+      );
+    }
+  }
+
+  async updateDependencies(
```
lukebatchelor: Personally I would either only expose `setDependencyVersionRange` *or* `updateDependencies`, not both.
There are cases where each would work.

I think you're right in that you'll quite often be editing one dep at a time, so I'm leaning towards setDependencyVersionRange. 

Happy either way though.

PR #28, opened 2017-09-22
Init exec commands

thejameskyle:
```
pyarn exec -- [command]
pyarn project exec -- [command]
pyarn workspace exec -- [command]
pyarn workspaces exec -- [command]
```


src/cli.js
```diff
@@ -394,7 +413,11 @@ export default async function cli(argv: Array<string>, exit: boolean = false) {
         workspace    run a pyarn command inside a specific workspace
         help         get help with pyarn commands
     `
-  });
+    },
+    {
+      '--': true
```
lukebatchelor: What's this for out of curiosity?

thejameskyle: It's what enables parsing of `--` inside minimist https://github.com/substack/minimist#var-argv--parseargsargs-opts

src/utils/execCommand.js
```diff
@@ -0,0 +1,33 @@
...
+
+  if (process.env.PATH) {
+    PATH_PARTS.push(process.env.PATH);
+  }
+
+  let PATH = PATH_PARTS.join(':');
+
+  return await processes.spawn(command, commandArgs, {
+    cwd: pkg.dir,
+    tty: false,
```
lukebatchelor: Do we want this to always be false? Maybe the commands should pass this in so that Project and Workspace (and maybe even workspaces if you filter it to contain one ws?) would get tty

lukebatchelor: Fine for now either way

lukebatchelor:
Looks sweet to me.

Have you done any manual testing of the main edge cases? (running in ws that doesnt have the bin, etc)

lukebatchelor:
The Close and comment button **should not** be so close to the comment button!

src/utils/execCommand.js
thejameskyle: I need to find a way to better support multiple processes logging at the same time.

I left it like this to revisit

thejameskyle:
This is ready to go now

PR #27, opened 2017-09-21
Pyarn add - take 2 - initial refactoring

lukebatchelor:
This is the initial refactoring for pyarn add.

It add yarn.add and symlinkPackageDependencies.

TODO: 
* symlinkPackage**s**Dependencies
* refactor install
* tests for install
* pyarn [w|ws|p] add + tests

PR #29, opened 2017-09-25
Turn config into a class with getters

thejameskyle:


src/Config.js
```diff
@@ -0,0 +1,212 @@
...
+    }
+  });
+}
+
+function toObjectOfStrings(value: JSONValue, message: string) {
+  if (typeof value !== 'object' || value === null || Array.isArray(value)) {
+    throw new PError(message);
+  }
+
+  let safeRef = value;
```
lukebatchelor: Is this actually required? I'm inclined to assume yes, but just wondering why? If we replaced safeRef with value everywhere here, I would think this would be just as safe?

lukebatchelor:
Sweet, all LGTM.

lukebatchelor:
Will rebase once this is merged. Will need to rebase #28 as well.

Would you prefer I rebase #27 with this and merge it, or just create one PR will all the rest of the Add stuff as well?

src/Config.js
thejameskyle: It avoids having a refinement invalidation. Flow assumes all functions could mutate values behind the scenes. Re-assigning the value means that I'm not changing the binding to be another value.

PR #31, opened 2017-09-26
Rework messages/logger to use opaque type, add emojis, rework running scripts to add message

thejameskyle:
You are now forced to write:

```js
logger.error(messages.didntWork())
```

This will cause a type error:

```js
logger.error(`Didn't work`);
// string. This type is incompatible with the expected param type of opaque type `Message`
```

I also added a bunch of emojis to the CLI, and I split out some logic from `yarn.run` to `yarn.getScript` so we could add messages for missing scripts elsewhere.

### DIFF
--- a/src/Project.js
+++ b/src/Project.js
@@ -181,7 +181,7 @@ export default class Project {
     );
 
     if (filteredWorkspaces.length === 0) {
-      logger.warn('No packages match the filters provided');
+      logger.warn(messages.noPackagesMatchFilters());
     }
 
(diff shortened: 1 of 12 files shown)


thejameskyle:
This is the output running `bolt install` now:

bolt-install

lukebatchelor:
Ha! 👌

Fantastic.

thejameskyle:
Goddamnit Luke

screen shot 2017-09-26 at 7 32 02 pm

PR #32, opened 2017-09-27
Add getProject function

d4rkr00t:


thejameskyle:
I think a `getProject` function that matches [`getWorkspaces`](https://github.com/boltpkg/bolt/blob/master/src/functions/getWorkspaces.js#L9-L13) would probably be better,

d4rkr00t:
Yeah, i can rename

thejameskyle:
Sorry, I wanted it both renamed and to return an object with `dir` `name` and `config` like `getWorkspaces` https://github.com/boltpkg/bolt/blob/master/src/functions/getWorkspaces.js#L9-L13

d4rkr00t:
Updated!

thejameskyle:
I changed the interface to be:

```js
bolt.getProject({ cwd: '/../' });
```

Where `cwd` defaults to `process.cwd()` like the other functions.

PR #34, opened 2017-10-04
Fixes workspace run command

lukebatchelor:
Just updates to use the yarn.getScript helper.

Added some tests and updated the example folder too.


thejameskyle:
thanks for writing tests

PR #35, opened 2017-10-06
Adds `bolt add` and `bolt workspace add` commands

lukebatchelor:
This is encompasses all the work from #27

* adds yarn.add helper
* adds symlinking helper
* refactors install to use symlinking
* adds tests for install

Its still a bit messy and theres some dupliacted code I'd like to clean up, but I wanna try to get this in and do any more refactors in the next drop (which would also add `bolt workspaces add`).

As an aside, the testing patterns I've been using have changed a fair bit over the course of all this, so I definitely want to refactor and align all that at some point.

The mock yarn.add command for instance is in three different forms in three different places. A single one that everything uses would be awesome.


lukebatchelor:
[] refactor nested-workspaces-with-root-dependencies-installed tests to use `package-with-external-deps-installed`

src/__fixtures__/nested-workspaces-with-scoped-package-names/packages/foo/package.json
```diff
@@ -6,6 +6,6 @@
   },
   "dependencies": {
     "react": "^15.6.1",
-    "bar": "^1.0.0"
+    "@scoped/bar": "^1.0.0"
```
lukebatchelor: Hmm... I'm going to double check all these fixures. I dont recall changing all these, so I'm thinking my tests may have modified them at some point.

- [ ] find out if I modified all these fixtures or not...

src/commands/__tests__/add.test.js
```diff
@@ -1,4 +1,179 @@
...
+  const dirExists = await pathExists(
+    path.join(workspaceDir, 'node_modules', depName)
+  );
+  const depInPkgJson = pkg.getDependencyType(depName) !== null;
+
+  return dirExists && depInPkgJson;
+}
+
+// a mock yarn add function to update the pakcages's config and also node_modules dir
+async function fakeYarnAdd(pkg: Package, dependencies, type = 'dependencies') {
```
lukebatchelor: This is the yarn.add mock I'll be using in the rest of the tests later. Just a much cleaner way of handling than what I was doing before.

src/commands/add.js
```diff
@@ -1,16 +1,56 @@
...
+import type { Dependency, configDependencyType } from '../types';
 
-export type AddOptions = {};
+export type AddOptions = {
+  cwd?: string,
+  deps: Array<Dependency>,
+  type: configDependencyType
+};
+
+const depTypeFlags = {
```
lukebatchelor: This and the toWorkspaceAddOptions are almost identical to the ones in add() bar the workspace arg. I'd love to dedupe if possible...

src/__fixtures__/nested-workspaces-with-scoped-package-names/packages/foo/package.json
lukebatchelor: Yup, all good.

lukebatchelor:
Okay, had another look through and this seems pretty fine. I'm going to start work on the addDependenciesToPackage **s** function now.

src/utils/yarn.js
```diff
@@ -1,28 +1,44 @@
...
     }
+  });
+
+  if (type) {
+    const typeToFlagMap = {
+      dependencies: '',
+      devDependencies: '--dev',
+      peerDependencies: '--peer',
+      optionalDependencies: '--optional'
+    };
```
thejameskyle: Might as well hoist this as a constant

src/utils/symlinkPackageDependencies.js
```diff
@@ -0,0 +1,219 @@
...
+  );
+
+  await Promise.all(
+    symlinksToCreate.map(async ({ src, dest, type }) => {
+      await fs.symlink(src, dest, type);
+    })
+  );
+
+  await yarn.run(pkg, 'postinstall');
+  await yarn.run(pkg, 'prepublish');
```
thejameskyle: Should also add `prepare`

- https://github.com/npm/npm/blob/latest/doc/misc/npm-scripts.md
- https://github.com/yarnpkg/yarn/pull/3004

src/utils/addDependenciesToPackages.js
```diff
@@ -0,0 +1,80 @@
...
+
+  for (let dep of externalDeps) {
+    const installed = project.pkg.getDependencyVersionRange(dep.name);
+    if (dep.version && dep.version !== installed) {
+      logger.warn(
+        messages.depMustMatchProject(
+          pkg.config.getName(),
+          dep.name,
+          installed,
+          String(dep.version)
```
thejameskyle: If you store the version in a variable, you won't need to check this again.


```jss
const installed = project.pkg.getDependencyVersionRange(dep.name);
const version = dep.version;
if (version && version !== installed) {
  logger.warn(
    messages.depMustMatchProject(
    pkg.config.getName(),
    dep.name,
    installed,
    version
  )
);
```

src/utils/addDependenciesToPackages.js
```diff
@@ -0,0 +1,80 @@
...
+        )
+      );
+      throw new Error();
```
thejameskyle: I'd rather throw the warning as an error instead of logging it and throwing an error.

src/utils/addDependenciesToPackages.js
```diff
@@ -0,0 +1,80 @@
...
+    if (dep.version) {
+      logger.warn(
+        messages.packageMustDependOnCurrentVersion(
+          pkg.config.getName(),
+          dep.name,
+          curVersion,
+          String(dep.version)
+        )
+      );
+      throw new Error();
```
thejameskyle: Same as above https://github.com/boltpkg/bolt/pull/35/files#r143371033

thejameskyle:
I think we should enforce at the project level that you cannot add a workspace as a dependency.

```sh
path/to/project/ $ bolt add workspace-name # error
path/to/project/other-workspace/ $ bolt add workspace-name # success

File: src/commands/workspace/__tests__/add.test.js [hunk 9]
```diff
@@ -1,4 +1,104 @@
...
+    expect(await depIsInstalled(fooWorkspaceDir, 'project-only-dep')).toEqual(
+      false
+    );
+    await workspaceAdd(
+      toWorkspaceAddOptions(['foo', 'project-only-dep'], {
+        cwd: barWorkspaceDir
+      })
+    );
+    expect(yarn.add).toHaveBeenCalledTimes(0);
+    expect(await depIsInstalled(fooWorkspaceDir, 'project-only-dep')).toEqual(
```
thejameskyle: btw `.toEqual` is meant for deep object comparison where `.toBe` is meant for reference equality. I generally use it for `.toBe(true)`-etc

thejameskyle:
Can you rebase?

src/commands/workspace/__tests__/add.test.js
lukebatchelor: Oh, no worries. I'll go  replace all of these in the next PR if you like

lukebatchelor:
I cant seem to push here after rebasing, so pushed to https://github.com/boltpkg/bolt/pull/36

PR #37, opened 2017-10-13
Adds yarn.runIfExists

lukebatchelor:
I introduced a regression with the bolt add when I refactored the install script to use my symlinkPakckages helper.

I've added `yarn.runIfExists` to wrap the behaviour where you dont want to error if the script doesnt exist. So, things like lifecycle events and `workspaces run`.

### DIFF
--- a/src/commands/workspaces/run.js
+++ b/src/commands/workspaces/run.js
@@ -34,6 +34,7 @@ export async function workspacesRun(opts: WorkspacesRunOptions) {
   );
 
   await Project.runWorkspaceTasks(filteredWorkspaces, async workspace => {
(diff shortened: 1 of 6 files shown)


src/commands/workspaces/run.js
```diff
@@ -34,6 +34,7 @@ export async function workspacesRun(opts: WorkspacesRunOptions) {
   );
 
   await Project.runWorkspaceTasks(filteredWorkspaces, async workspace => {
-    await yarn.run(workspace.pkg, opts.script, opts.scriptArgs);
```
lukebatchelor: This one had no check, I figure it should be a non-error case either way

PR #42, opened 2017-10-26
Fixes bug when updating multiple package deps at once

lukebatchelor:


.gitignore
```diff
@@ -1,5 +1,5 @@
 node_modules
 *.log
 dist
-coverage
```
lukebatchelor: Weird, this actually works for ignoring the directory in git, but vscode wouldnt recognise it (I'd still get search results from this dir).

src/__tests__/Config.test.js
```diff
@@ -5,6 +5,8 @@ import { mkdtempSync } from 'fs';
 import * as path from 'path';
 import { getFixturePath } from 'jest-fixtures';
 
+jest.spyOn(console, 'error').mockImplementation(() => {});
```
thejameskyle: Why mock this instead of `logger`?

lukebatchelor: ... ... I Completely forgot which repo I was in. :man_facepalming:

PR #47, opened 2017-10-31
Temporarily disables locking during publish

lukebatchelor:
Will release a 0.15.0 for this

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bolt",
-  "version": "0.14.3",
+  "version": "0.14.4",
   "main": "index.js",
   "author": "James Kyle <me@thejameskyle.com>",
   "license": "MIT",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "bolt",
-  "version": "0.14.2",
```
lukebatchelor: What I think happened: When we were releasing on my machine yesterday we published 0.14.3 and 0.14.4, but had deleted the tag for 0.14.4. I had done a bunch more local work without pushing the changes back, including manually changing the version number so I could push to a forked, scoped bolt.

Long story short: 0.14.3 and 0.14.4 are both published, but the 0.14.4 tag looks like it was missing.

I've just released 0.14.5 (with these changes) and the tag is there correctly.

PR #49, opened 2017-10-31
Adds validateProject util

lukebatchelor:
Prevents installing in a repo where the project has a dep on a workspace.


lukebatchelor:
Agh. I rebased this on the other PR because at one point I was checking the depType. 

I dont need that anymore, so I'll repush to another PR

PR #52, opened 2017-11-01
Moves bolt engine check to bolt.version instead of engines.bolt

lukebatchelor:
Due to warning from yarn when installing

```
⚡️   bolt v0.15.1 (node v8.4.0)
🔎   [1/3] Validating project...
📦   [2/3] Installing project dependencies...
$ yarn install --non-interactive -s warning @atlaskit/fabric@0.0.0: The engine "bolt" appears to be invalid.
```

See https://github.com/yarnpkg/yarn/blob/master/src/package-compatibility.js#L13 and 
https://github.com/yarnpkg/yarn/blob/master/src/package-compatibility.js#L48 to see why it breaks (we'd need to either be able to write to the process.engines field (I tried) or have bolt hardcoded into yarn).


lukebatchelor:
@thejameskyle gonna merge this today because we'll be making a breaking change later and I'd like to be able to warn people

PR #44, opened 2017-10-29
switch to using semver to compare matching versions

Noviny:
@lukebatchelor if you could have a look at this, that would be great.

PR #55, opened 2017-11-02
Allow internal workspace dependencies to semver-satisfy rather than equal

lukebatchelor:
This is just finishing @Noviny 's work on #44 .

We realised whilst testing this that it can lead to issues if the external dependencies arent kept strictly equal.

I've just reverted the external deps changes and added tests for the internal deps changes.

The point of this PR is to allow a package to have a caret dependency on a workspace package and not have to update its deps every time that dependency is bumped. 

Once this is merged, we can update build-releases to only bump deps if they no longer semver-satisfy.

### DIFF
--- a/src/utils/__tests__/addDependenciesToPackages.test.js
+++ b/src/utils/__tests__/addDependenciesToPackages.test.js
@@ -95,6 +95,22 @@ describe('utils/addDependenciesToPackages', () => {
     ).rejects.toBeInstanceOf(Error);
   });
 
+  test('it should accept a semver range that satisfies the project config ', async () => {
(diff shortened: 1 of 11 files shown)


src/utils/__tests__/addDependenciesToPackages.test.js
```diff
@@ -137,9 +137,27 @@ describe('utils/addDependenciesToPackages', () => {
     );
   });
 
+  test('should be able to add packages with tagged versions (without specifying)', async () => {
```
lukebatchelor: This would have always worked, but decided to add a defensive test just in case.

thejameskyle:
Fix your git config

lukebatchelor:
What do you mean? o.O

PR #58, opened 2017-11-08
Updates bolt.updatePackageVersions to not automatically add a caret

lukebatchelor:
As in title. This is to allow us to bump internal versions only when required.

Note: This is a breaking change from current API.

### DIFF
--- a/src/functions/__tests__/updatePackageVersions.test.js
+++ b/src/functions/__tests__/updatePackageVersions.test.js
@@ -33,7 +33,7 @@ describe('function/updatePackageVersions', () => {
       expect(await getDepVersion(barPath, 'react')).toBe('^15.6.1');
       expect(await getDepVersion(bazPath, 'react')).toBe('^15.6.1');
 
-      await updatePackageVersions({ react: '15.6.0' }, { cwd });
+      await updatePackageVersions({ react: '^15.6.0' }, { cwd });
 
       expect(await getDepVersion(fooPath, 'react')).toBe('^15.6.0');
       expect(await getDepVersion(barPath, 'react')).toBe('^15.6.0');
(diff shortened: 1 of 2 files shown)


lukebatchelor:
Hmm... This has now got me thinking...
If we just use this, then the logic for when to bump an internal dep lives in the other repo (which would soon be bolt-releases, so maybe not as bad). It feels like it belongs in bolt.

I had a look at checking if its an internal or external dep in the updatePackageVersions.js function, but that only works if you are being passes explicit versions and would remove the ability to set internal versions to some sort of range.

So, maybe what would make more sense, would be to create a separate updatePackageVersions function that just looks at all internal dependencies and decides if they need to be bumped in order to keep the repo in sync. Maybe it even takes a list of expected packages and if bolt finds one that needs to be bumped that you havent mentioned, it can throw.

So, something like (in ak-mk-2):

```js
const releaseObj = createRelease(unreleasedChangesets, allPackages);
const publishCommit = createReleaseCommit(releaseObj);
// ...
await bumpReleasedPackages(releaseObj, allPackages);
// Need to transform releases into a form for bolt to update dependencies
const versionsToUpdate = releaseObj.releases.reduce((cur, next) => ({
 ...cur,
   [next.name]: next.version,
  }), {});
 // update dependencies on those versions
 await bolt.updateInternalPackageVersions(versionsToUpdate);
```

I'll think about it tonight and will implement whichever solution will be easiest to revert without breaking things.

lukebatchelor:
So, after thinking about it more, I think I'm fine to have that logic live in bolt releases. The two places we'd need it in our build is in the changeset command (to find which dependents we need to prompt for) and in release (to know which dependents to bump). Both of which will be in bolt-releases, so I think its fine.

Issue #59, opened 2017-11-14
Publish should warn when a package is not releases due to later version being on npm


lukebatchelor:
Currently, we just silently continue.

Not sure if it should be a separate issue or not, but as a part of this, I'd also like publish to return a list of published packages. This will help solve an issue we have currently where we report that everything was published as expected, even when some fail.

ajaymathur:
PR merged.

PR #60, opened 2017-11-15
Amathur/issue#59

ajaymathur:
1. make server.gt show warning if failure - **done**
2. return actual list of packages published - **WIP**

Fix for https://github.com/boltpkg/bolt/issues/59

### DIFF
--- a/src/commands/publish.js
+++ b/src/commands/publish.js
@@ -24,6 +24,20 @@ export function toPublishOptions(
 }
 
 async function getUnpublishedPackages(packages) {
+  const semverGtCheckFailWithWarning = (
+    pkgLocalVersion: string | number,
+    pkgPublishedVersion: string | number,
+    pkgName: string
+  ): boolean =>
+    semver.gt(pkgLocalVersion, pkgPublishedVersion) ||
+    !!logger.warn(
+      messages.notPublishingPackage(
+        pkgLocalVersion,
+        pkgPublishedVersion,
+        pkgName
+      )
+    );
+
   const results = await Promise.all(
     packages.map(async pkg => {
       let config = pkg.config;
@@ -42,7 +56,11 @@ async function getUnpublishedPackages(packages) {
     // only publish if our version is higher than the one on npm
     return (
       !result.isPublished ||
-      semver.gt(result.localVersion, result.publishedVersion)
+      semverGtCheckFailWithWarning(
+        result.localVersion,
+        result.publishedVersion,
+        result.name
+      )
     );
   });
 }
--- a/src/commands/publish.js
+++ b/src/commands/publish.js
@@ -25,18 +25,25 @@ export function toPublishOptions(
 
 async function getUnpublishedPackages(packages) {
   const semverGtCheckFailWithWarning = (
-    pkgLocalVersion: string | number,
-    pkgPublishedVersion: string | number,
+    pkgLocalVersion: string,
+    pkgPublishedVersion: string,
     pkgName: string
-  ): boolean =>
-    semver.gt(pkgLocalVersion, pkgPublishedVersion) ||
-    !!logger.warn(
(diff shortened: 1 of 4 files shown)


thejameskyle:
@ajaymathur You probably want to update your git config to use your GitHub email, otherwise these commits won't be tied to your GitHub account.

thejameskyle:
@lukebatchelor can you take this?

lukebatchelor:
Yup

ajaymathur:
@lukebatchelor : ready for review 📃

src/commands/publish.js
```diff
@@ -74,9 +100,18 @@ export async function publish(opts: PublishOptions) {
       const version = workspace.pkg.config.getVersion();
       logger.info(messages.publishingPackage(name, version));
 
-      await npm.publish(name, { cwd: workspace.pkg.dir, access: opts.access });
+      const publishConfirmation = await npm.publish(name, {
+        cwd: workspace.pkg.dir,
+        access: opts.access
+      });
+
+      if (publishConfirmation && publishConfirmation.published) {
```
ajaymathur: this is so that test case passes

src/utils/npm.js
```diff
@@ -39,10 +39,24 @@ export function publish(
...
+
+    try {
+      const result = await processes.spawn(
+        'npm',
+        ['publish', ...publishFlags],
+        {
+          cwd: opts.cwd
+        }
+      );
+      return { published: true, publishedPkgInfo: JSON.parse(result.stdout) };
```
lukebatchelor: Not sure about the naming of this one. 

From my understanding, this will return something like

```
$ node-client git:(patch-1) ✗ npm publish
+ @lbatchelor/netlify@1.2.0-patched.6
```

Which wouldnt be parseable.

I would just return true or false from this I think. The error messages should be written to stderr already, so no need for us to capture that.

src/utils/npm.js
```diff
@@ -39,10 +39,24 @@ export function publish(
...
+    } catch (error) {
+      //Publish failed
+      if (error.code == 0) {
```
lukebatchelor: I would remove this check too, and just return false (Not sure if you can return false directly, might need to wrap it in a promise 

```
return Promise.resolve(true);
```

We can check tomorrow

src/commands/publish.js
```diff
@@ -24,6 +24,27 @@ export function toPublishOptions(
...
 
 async function getUnpublishedPackages(packages) {
+  const semverGtCheckFailWithWarning = (
+    pkgLocalVersion: string,
+    pkgPublishedVersion: string,
+    pkgName: string
+  ): boolean => {
+    const shouldPublish = semver.gt(pkgLocalVersion, pkgPublishedVersion);
+
+    //show a warning message if package is not published since pkgPublishedVersion > pkgLocalVersion
```
lukebatchelor: nit: could we add a space here

```
// show a ...
```

src/commands/publish.js
```diff
@@ -74,9 +100,18 @@ export async function publish(opts: PublishOptions) {
...
+        cwd: workspace.pkg.dir,
+        access: opts.access
+      });
+
+      if (publishConfirmation && publishConfirmation.published) {
+        publishedPackages.push({ name, version });
+      }
     });
 
+    return publishedPackages;
```
lukebatchelor: :partyparrot:

ajaymathur:
added tests, though I am not sure if it is the accepted pattern, 😕

lukebatchelor:
Nope, test looks great. 

I actually meant a test for the returned packages though. 

Something where we use mocks to make it go to publish 3 packages, have two of the succeed (published: true) and one fail, and check that only the two packages get returned.

lukebatchelor:
Sweet, all tested manually. Awesome first PR, lets get this into the build today too.

PR #61, opened 2017-11-15
Changes updatePackageVersions to work with ranges

lukebatchelor:
* updatePackageVersions now supports bumping packages with caret and tilde ranges
* BREAKING: it will no longer support bumping external deps and will warn if you attempt to

We could rename this at some point, but I think this greatly simplifies the functions purpose:
"An external tool has bumped your versions, run this function to get all your internal versions back in to line"

It should be noted though, it only bumps the ones you've told it to, and doesnt perform any validation of what you've passed in, besides excluding  external deps.

### DIFF
--- a/src/functions/updatePackageVersions.js
+++ b/src/functions/updatePackageVersions.js
@@ -1,6 +1,8 @@
 // @flow
 import Project from '../Project';
 import Workspace from '../Workspace';
+import * as logger from '../utils/logger';
+import * as messages from '../utils/messages';
 
 type VersionMap = {
   [name: string]: string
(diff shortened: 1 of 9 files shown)


src/functions/updatePackageVersions.js
```diff
@@ -10,8 +12,23 @@ type Options = {
   cwd?: string
 };
 
+function versionRangeToRangeType(versionRange: string) {
```
Noviny: I'd love this function to be more certain and check the whole regex is correct. `/^\^\d+\.\d+\.\d+$/` for example. It means we can throw/warn when a range someone provided is overwritten (for example they provided `^1.5.0 || ^2.0.0`)

src/functions/updatePackageVersions.js
```diff
@@ -20,20 +37,33 @@ export default async function updatePackageVersions(
...
+      const depTypes = pkg.getDependencyTypes(depName);
+      const rangeType = versionRangeToRangeType(depRange);
+
       if (depTypes.length === 0) continue;
       for (let depType of depTypes) {
         await pkg.setDependencyVersionRange(
           depName,
           depType,
-          versions[depName]
+          rangeType + versions[depName]
```
Noviny: Not a huge fan of doing actions inside a function call's arguments.

src/functions/updatePackageVersions.js
lukebatchelor: I think I'm happy to leave it for now, because that check will be a part of bolts project validation soon.
Same as the check for, do your peerDep and devDeps match. This function assumes they do. We dont currently enforce that, but should soon.

lukebatchelor: Raised https://github.com/boltpkg/bolt/issues/62

src/functions/updatePackageVersions.js
lukebatchelor: agreen

PR #63, opened 2017-11-15
Pass pkg to processes.spawn from execCommand

d4rkr00t:
Before:
```sh
$ tsc --project ./build/es5 src/ui/MediaItem/MediaComponent.tsx(237,27): error TS2339: Property 'src' does not exist on type 'Blob'.
```

After: 
```sh
(@atlaskit/editor-common) $ tsc --project ./build/es5 src/ui/MediaItem/MediaComponent.tsx(237,27): error TS2339: Property 'src' does not exist on type 'Blob'.
```

### DIFF
--- a/src/utils/execCommand.js
+++ b/src/utils/execCommand.js
@@ -26,6 +26,7 @@ export default async function execCommand(
   let PATH = PATH_PARTS.join(':');
 
   return await processes.spawn(command, commandArgs, {
+    pkg,
     cwd: pkg.dir,
     tty: false,
     env: { ...process.env, PATH }


lukebatchelor:
Hmmm... This is exposing Package to external consumers right? ( If I understand correctly here).

I think we'd want to avoid that. 

Your description looks like you've copied the same message in both places except the prompt string? (Is that what this was for)

d4rkr00t:
Yeah, it just passes pkg to processes.spawn so it can prefix messages with the package name, otherwise it's very hard to tell from which packages the output is coming from

d4rkr00t:
And it's not exposing Package in any way to external consumers

PR #66, opened 2017-11-17
docs update added bolt bin in commands

ajaymathur:
added `bolt bin` in commands + update formating


lukebatchelor:
Sweet! Nice catch

Issue #56, opened 2017-11-08
bolt add --dev pkgName does not seem to work


lukebatchelor:
When running

```
bolt add --dev pkgA pkgB pkgC
```

pkgB and pkgC will be added correctly, but pkgA will not.

After some investigation, we think that meow might be taking pkgA and assigning that as the value of the --dev flag, which will still be truthy, so everything else will work as expected.

PR #67, opened 2017-11-17
[fix#56] picking value of --<dependencyType> flag and updating depend…

ajaymathur:
…cies Fixes https://github.com/boltpkg/bolt/issues/56

### DIFF
--- a/src/commands/add.js
+++ b/src/commands/add.js
@@ -41,6 +41,10 @@ export function toAddOptions(
   Object.keys(DEPENDENCY_TYPE_FLAGS_MAP).forEach(depTypeFlag => {
     if (flags[depTypeFlag]) {
       type = DEPENDENCY_TYPE_FLAGS_MAP[depTypeFlag];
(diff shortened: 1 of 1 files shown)


src/commands/add.js
```diff
@@ -41,6 +41,10 @@ export function toAddOptions(
   Object.keys(DEPENDENCY_TYPE_FLAGS_MAP).forEach(depTypeFlag => {
     if (flags[depTypeFlag]) {
       type = DEPENDENCY_TYPE_FLAGS_MAP[depTypeFlag];
+      // check if value of dependency flag is a package name and then push to dependency arguments
+      if (typeof flags[depTypeFlag] === 'string') {
```
ajaymathur: kept it this way instead of creating const - so that if there are any updates in usage of meow this code can be removed with any further refactor

lukebatchelor: Ah, nice solution.

Issue #53, opened 2017-11-02
Supported yarn version?


lukebatchelor:
I just tried running bolt on a different machine with yarn 0.22.x installed and it fails to run any lifecycle events.

Looks like

```
yarn run -s script
```

Doesnt work properly in older yarns. Do we want to support that?

We can make it

```
yarn run script -s -- other args
```

And that should be compatible with both.

treshugart:
It would be great to do one of the following:

1. Add `yarn` as a `peerDependency` so consumers at least know they need it and which version they need. I and others have had version mismatch issues but didn't know it at first until digging really deep or hail-mary-updating `yarn`.
2. Install yarn as a `dependency` and whenever we call out to it, use the local version instead of a globally installed version.

I've talked to @thejameskyle about option 2 in the past, and I think that'd be the preferred option if nothing is blocking it.

jameslnewell:
Was also on `0.22.x` and putting up with a heap of weirdness

thejameskyle:
> Install yarn as a dependency and whenever we call out to it, use the local version instead of a globally installed version.

We should do this sooner rather than later

PR #69, opened 2017-11-23
added yarn as dependency and running locally installed yarn

ajaymathur:
Fixes https://github.com/boltpkg/bolt/issues/53

### DIFF
--- a/src/utils/yarn.js
+++ b/src/utils/yarn.js
@@ -3,10 +3,16 @@ import includes from 'array-includes';
 
 import type { Dependency, configDependencyType } from '../types';
 import type Package from '../Package';
+import Project from '../Project';
 import * as processes from './processes';
 import * as fs from '../utils/fs';
 import * as logger from '../utils/fs';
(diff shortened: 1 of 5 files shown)


thejameskyle:
Why peerDependency?

ajaymathur:
yeah, it should be dependency. Updated 🙂

lukebatchelor:
Sweet, those tests were useful to update.

src/utils/yarn.js
```diff
@@ -3,11 +3,17 @@ import includes from 'array-includes';
...
 import type Package from '../Package';
+import Project from '../Project';
 import * as processes from './processes';
 import * as fs from '../utils/fs';
 import * as logger from '../utils/fs';
 import { DEPENDENCY_TYPE_FLAGS_MAP } from '../constants';
 
+async function getProjectBinPath(cwd: string): Promise<string> {
+  const project = await Project.init(cwd);
+  return project.pkg.nodeModulesBin;
```
thejameskyle: Had to revert this in #70 because this should be pulling from the bin path of `bolt` not the project.

PR #72, opened 2017-11-24
added yarn as dependency and running local yarn for all yarn related …

ajaymathur:
…bolt commands

Fixes https://github.com/boltpkg/bolt/issues/53


ajaymathur:
*** Fixed the same thing in https://github.com/boltpkg/bolt/pull/73 , I think that is better with coding style but I had to add a new package published https://yarnpkg.com/en/package/project-bin-path , can discuss which is better implementation***

PR #75, opened 2017-11-29
publish return relevant data

Noviny:


lukebatchelor:
Sweet! LGTM

lukebatchelor:
Reckon we could/should explicitly state the return value shape of publish?

src/commands/publish.js
```diff
@@ -106,7 +105,17 @@ export async function publish(opts: PublishOptions) {
       });
 
       if (publishConfirmation && publishConfirmation.published) {
-        publishedPackages.push({ name, version });
+        publishedPackages.push({
```
lukebatchelor: Could this be rewritten as

```
publishedPackages.push({
           name,
           newVersion: version,
           published: publishConfirmation && publishConfirmation.published
         });
```

Noviny: Yes. Good call.

PR #77, opened 2017-11-30
removed condition to check for subcommand being undefined in case of …

ajaymathur:
…workspace run


lukebatchelor:
Sweet. Same as the other. LGTM, lets merge next week.

PR #79, opened 2017-12-01
[WIP] Workspace link unlink commands

ajaymathur:
[WIP] 🎢

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -84,8 +84,10 @@ const commandMap = {
   WORKSPACE: { workspace: true, w: true },
   WORKSPACE_ADD: { add: true },
   WORKSPACE_EXEC: { exec: true },
+  WORKSPACE_Link: { link: true },
   WORKSPACE_REMOVE: { remove: true, rm: true },
   WORKSPACE_RUN: { run: true },
+  WORKSPACE_UNLINK: { unlink: true },
   WORKSPACE_UPGRADE: { upgrade: true },
   WORKSPACES: { workspaces: true, ws: true },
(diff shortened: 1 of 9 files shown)


src/commands/workspace/__tests__/link.test.js
```diff
@@ -0,0 +1,72 @@
+//@flow
+import { workspaceLink } from '../link';
+import { copyFixtureIntoTempDir } from 'jest-fixtures';
```
ajaymathur: @lukebatchelor need help with test , 🆘

src/cli.js
```diff
@@ -84,8 +84,10 @@ const commandMap = {
   WORKSPACE: { workspace: true, w: true },
   WORKSPACE_ADD: { add: true },
   WORKSPACE_EXEC: { exec: true },
+  WORKSPACE_Link: { link: true },
```
lukebatchelor: `WORKSPACE_LINK`*

lukebatchelor:
Hmm... So, linking and unlinking should only really apply to external deps right?
Internal deps are already linked and should always remain that way right?

* Attempting to link an internal dep should warn/throw
* Attempting to unlink an internal dep should throw
* Linking an external dep should link at the root level (running yarn link there) then checking that the required symlinks in the packages still exist. (They should already be pointing to the root).
* Unlinking should be the same
* There is an error/warning when symlinking that warns when packages are linked outside the repo, we might need to update that message

ajaymathur:
closing this in favour of https://github.com/boltpkg/bolt/pull/81

PR #80, opened 2017-12-05
[WIP] Run tasks with graph-sequencer

thejameskyle:
Built this library today: https://github.com/thejameskyle/graph-sequencer

It gives us chunks to run in topological with cycles resolved via the workspaces config.

I still have a bunch of work left to do but this is the meat of it.


lukebatchelor:
LGTM

PR #81, opened 2017-12-21
bolt 🔗 and !🔗 command

ajaymathur:
Test cases are pending, and minor refactor. but ready for initial review

### DIFF
--- a/src/commands/link.js
+++ b/src/commands/link.js
@@ -1,16 +1,52 @@
 // @flow
 import * as options from '../utils/options';
+import * as yarn from '../utils/yarn';
 import { BoltError } from '../utils/errors';
+import Project from '../Project';
 
-export type LinkOptions = {};
+function getWorkspaceMap(workspaces) {
+  let workspaceMap = new Map();
+
+  for (let workspace of workspaces) {
+    workspaceMap.set(workspace.pkg.config.getName(), workspace);
+  }
+
+  return workspaceMap;
+}
+
+export type LinkOptions = {|
+  cwd?: string,
+  packagesToLink: ?Array<string>
+|};
 
 export function toLinkOptions(
   args: options.Args,
   flags: options.Flags
 ): LinkOptions {
-  return {};
+  return {
+    cwd: options.string(flags.cwd, 'cwd'),
+    packagesToLink: args
+  };
 }
 
 export async function link(opts: LinkOptions) {
-  throw new BoltError('Unimplemented command "link"');
+  let cwd = opts.cwd || process.cwd();
+  let packagesToLink = opts.packagesToLink;
+  let project = await Project.init(cwd);
+  let workspaces = await project.getWorkspaces();
+  let workspaceMap = getWorkspaceMap(workspaces);
+
+  if (packagesToLink && packagesToLink.length) {
+    packagesToLink.forEach(async packageToLink => {
+      if (workspaceMap.has(packageToLink)) {
+        throw new BoltError(
+          `There is a workspace by the name ${packageToLink} already linked`
+        );
+      }
+
+      await yarn.link(cwd, packageToLink);
+    });
+  } else {
+    throw new BoltError(`Need to pass project that needs to be linked`);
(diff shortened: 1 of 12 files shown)


ajaymathur:
build this from ground up keeping @lukebatchelor comment:

> Hmm... So, linking and unlinking should only really apply to external deps right?
> Internal deps are already linked and should always remain that way right?
> 
> Attempting to link an internal dep should warn/throw
> Attempting to unlink an internal dep should throw
> Linking an external dep should link at the root level (running yarn link there) then checking that the required symlinks in the packages still exist. (They should already be pointing to the root).
> Unlinking should be the same
> There is an error/warning when symlinking that warns when packages are linked outside the repo, we might need to update that message

ajaymathur:
PR is ready 💯 

Scenarios covered:

1. `bolt link foo`, ( where is foo is project workspace)
 => warning internal workspace cannot be linked. 🔴 

2. `bolt link some-external-pacakge`
=> links package in root 👍 

3. `bolt link [..multiple-pakages]` 
=> warning for all internal workspace and link all the external packages 👍 

---

On workspace 

`bolt w foo link`
=> will successfully create a link to foo

`bolt w foo link bar`, bar is an internal package
=> will follow scenario 1

`bolt w foo link some-external dependency`
=> will follow scenario 2

---

_same for unlink  🙂_

src/commands/link.js
```diff
@@ -1,16 +1,54 @@
...
+  let workspaceMap = getWorkspaceMap(workspaces);
+
+  if (packagesToLink && packagesToLink.length) {
+    packagesToLink.forEach(async packageToLink => {
+      if (workspaceMap.has(packageToLink)) {
+        logger.warn(messages.linkInternalPackage(packageToLink));
+      } else {
+        await yarn.link(cwd, packageToLink);
+      }
+    });
```
thejameskyle: Can you write it like this?

```js
await Promise.all(packagesToLink.map(async packageToLink => {
  // ...
}));
```

Right now `link()` will resolve before all the packages are linked because nothing is `await`-ing the promises being created inside the `forEach`

src/commands/unlink.js
```diff
@@ -1,16 +1,57 @@
...
+        if (workspaceMap.has(packageToUnlink)) {
+          logger.warn(messages.unlinkInternalPackage(packageToUnlink));
+        } else {
+          await yarn.unlink(cwd, packageToUnlink);
+        }
+      })
+    );
+  } else {
+    throw new BoltError(
+      'Please specify package to unlink. Try: bolt w [packageToUnlink] unlink'
```
thejameskyle: `bolt workspace [dep] unlink`

thejameskyle: Actually, why scoped under `workspace/w`?

src/commands/workspace/unlink.js
```diff
@@ -0,0 +1,52 @@
+// @flow
```
thejameskyle: I'm confused why there is a `bolt workspace [package] unlink` command at all

src/commands/workspace/link.js
```diff
@@ -0,0 +1,50 @@
+// @flow
```
thejameskyle: Same for `bolt workspace [package] link`

ajaymathur: Is helpful during migration.  
If packages in other repo are dependent on packages bolt repo.

src/commands/unlink.js
ajaymathur: So that we can create link only to the packages/ workspaces instead of complete project,
since other project/ repos will be dependent only on a packages. is my understanding correct?

thejameskyle: It's weird to me that you wouldn't want to link it in every package.

If packages A B and C all depend on X and you run `bolt workspace C link X` it seems odd that you wouldn't want to be linking it everywhere

ajaymathur: it will get linked everywhere in the project.

Issue #76, opened 2017-11-29
"up" as an alias to "upgrade"


thejameskyle:
I think it would be nice to be able to run:

```js
bolt up [dep]
bolt workspaces up [dep]
...
```

PR #83, opened 2017-12-26
up as an alias to upgrade

rajzshkr:
Fixes https://github.com/boltpkg/bolt/issues/76


ajaymathur:
sweet and thanks for the PR @rajzshkr  ❤️ .
Need to make changes for workspace upgrade and workspaces upgrade too 🙂

rajzshkr:
I missed in few other places also. Updated the PR.

PR #85, opened 2017-12-26
bolt info implementation

rajzshkr:
- bolt info package-name
- bolt info package-name --json

### DIFF
--- a/src/commands/info.js
+++ b/src/commands/info.js
@@ -8,19 +8,30 @@ export type InfoOptions = {|
   args: options.Args
 |};
 
+function flagsToArgs(flagArgs: options.Flags): Array<string> {
+  const flagArgsKey = Object.keys(flagArgs);
+  // Flags has default -- key
+  if (flagArgsKey.length < 1) {
+    return [];
+  }
+  const flag = [];
+  flagArgsKey.forEach(key => {
+    if (flagArgs[key] === true) {
+      flag.push(`--${key}`);
+    }
+  });
+  return flag;
+}
+
 export function toInfoOptions(
   args: options.Args,
   flags: options.Flags
 ): InfoOptions {
-  const depsArgs = Array.prototype.concat([], args);
-  Object.keys(flags).forEach(key => {
-    if (flags[key] === true) {
-      depsArgs.push(`--${key}`);
-    }
-  });
+  const infoFlagArgs = flagsToArgs(flags);
+  const infoSpawnArgs = Array.prototype.concat([], args, infoFlagArgs);
   return {
     cwd: options.string(flags.cwd, 'cwd'),
-    args: depsArgs
+    args: infoSpawnArgs
   };
 }
 
--- a/src/commands/info.js
+++ b/src/commands/info.js
@@ -5,8 +5,7 @@ import * as yarn from '../utils/yarn';
 
 export type InfoOptions = {|
   cwd?: string,
-  args: options.Args,
-  flags?: options.Flags
+  args: options.Args
 |};
 
 export function toInfoOptions(
--- a/src/commands/info.js
+++ b/src/commands/info.js
@@ -1,16 +1,30 @@
 // @flow
 import * as options from '../utils/options';
 import { BoltError } from '../utils/errors';
+import * as yarn from '../utils/yarn';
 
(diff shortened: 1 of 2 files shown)


src/utils/yarn.js
```diff
@@ -122,3 +122,11 @@ export async function bin(cwd: string) {
     tty: true
   });
 }
+
+export async function info(cwd: string, dependencies: Array<string> = []) {
```
thejameskyle: `dependencies`?

The arguments to `yarn info` are `yarn info [pkg] <key>`:

```js
$ yarn info lodash
{ name: 'lodash',
  description: 'Lodash modular utilities.',
  'dist-tags': { latest: '4.17.4' },
  versions:
   [ '0.1.0',
     '0.2.0',
     '0.2.1',
     '0.2.2',
     '0.3.0',
     '0.3.1',
     '0.3.2',
     '0.4.0',
     '0.4.1'
     ... }
$ yarn info lodash versions
[ '0.1.0',
  '0.2.0',
  '0.2.1',
  '0.2.2',
  '0.3.0',
  '0.3.1',
  '0.3.2',
  '0.4.0',
  '0.4.1',
  ... ]
```

We should make the signature `info(cwd: string, pkg: string, key?: string)`

src/commands/info.js
```diff
@@ -1,16 +1,34 @@
...
 
 export function toInfoOptions(
   args: options.Args,
   flags: options.Flags
 ): InfoOptions {
-  return {};
+  const depsArgs = Array.prototype.concat([], args);
+  Object.keys(flags).forEach(key => {
+    if (flags[key] === true) {
+      depsArgs.push(`--${key}`);
```
thejameskyle: We shouldn't convert flags like this. We should explicitly support individual flags

src/utils/yarn.js
rajzshkr: We've to send an array of arguments to the spawn process (https://github.com/boltpkg/bolt/blob/master/src/utils/processes.js#L67). Renamed to `dependencies` to `spawnArgs`.

src/commands/info.js
rajzshkr: > We should explicitly support individual flags

I didn't understand this.  I've seen the same kind of implementation in the project https://github.com/boltpkg/bolt/blob/master/src/utils/yarn.js#L16.

rajzshkr:
@thejameskyle Updated the PR. I' recently started work on boltpkg, I take reference from the project to implement new feature. If something is not correct, please give your comments. Always happy to make changes.

src/commands/info.js
thejameskyle: That's a very explicit conversion. I'm saying we shouldn't accept arbitrary flags and pass them through

src/commands/info.js
```diff
@@ -1,16 +1,45 @@
...
-export type InfoOptions = {};
+export type InfoOptions = {|
+  cwd?: string,
+  args: options.Args
+|};
+
+function flagsToArgs(flagArgs: options.Flags): Array<string> {
+  const flagArgsKey = Object.keys(flagArgs);
+  // Flags has default -- key
+  if (flagArgsKey.length < 1) {
```
ajaymathur: It should be `<= 1`  🙂

src/commands/info.js
```diff
@@ -1,16 +1,45 @@
...
+    return [];
+  }
+  const flag = [];
+  flagArgsKey.forEach(key => {
+    if (flagArgs[key] === true) {
+      flag.push(`--${key}`);
```
ajaymathur: flags received here are converted to camelCase by meow, therefore, they may not pass down to yarn same as what bolt receives, if json is the only flag supported (AFAIK) we can capture it explicitly.

ajaymathur: flags received here are converted to camelCase by meow, therefore, they may not pass down to yarn same as what bolt receives, if json is the only flag supported (AFAIK) we can capture it explicitly.

src/commands/info.js
rajzshkr: @ajaymathur I removed this piece of code.

src/commands/info.js
rajzshkr: Done !!

Issue #90, opened 2017-12-28
pass on the -- flags to scripts


ajaymathur:
this enable to pass options to scripts from cli 🙂

bradleyayers:
Not really sure what this issue is about, but I'm hoping it's related. I basically want to run `bolt w @heydovetail/dql test --watch`, where my test script is just running Jest. Is there a way to do that today without writing another NPM script (e.g. `test:watch`)?

lukebatchelor:
Not currently, no. But yes, I believe that's what this issue was for.

I can't remember exactly where this ended up, but I think the complexity was around how we know which flags are supposed to be passed to the script, and which are passed to bolt itself.

i.e

```
bolt ws run test --only="@heydovetail/dql" --watch
```

The only is meant for bolt, the watch is meant for the script.

d4rkr00t:
That's why I like npm's approach more then yarn's. Because in npm it's clear which arguments belongs where :)

lukebatchelor:
How's that @d4rkr00t ? using an extra `--`? That's what we do for separating the command when running exec

```
bolt ws exec --only="@heydovetail/dql" -- yarn test --watch
```

d4rkr00t:
Yeah i like `--` as it's clearer to me, that it separates npm flags from command flags...

lukebatchelor:
Yea, I agree actually. If @jamiebuilds agrees, I could look at fixing that soon™?

marcins:
I put some examples of a related issue we ran into in #183. FWIW running a yarn command with `--` gives a deprecation warning:

```
› yarn lint:css -- --bamboo
yarn run v1.9.2
warning From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.
```

bradleyayers:
I'm a 👎🏻to using `--`. It's more explicit which is great, but for me that's not enough to outweigh the pain of typing it. Explicit is great for programming languages where there are more reads than writes, but for a CLI I think ergonomics win. For example it's why I type `-h` instead of `--help`.

lukebatchelor:
The only other way to do it then would be to have the order matter, and I think even with that, it would still cause problems for any tool that took any of the same flags we want to use (i.e meow would either only accept one, or possibly treat it as an array).

Personally, I'd still be for the `--`. I'll check with Jamie tomorrow.

bradleyayers:
[...]

PR #93, opened 2017-12-31
pass the flags to scripts

ajaymathur:
Reference https://github.com/boltpkg/bolt/issues/90

### DIFF
--- a/src/commands/run.js
+++ b/src/commands/run.js
@@ -8,7 +8,8 @@ import { BoltError } from '../utils/errors';
 export type RunOptions = {|
   cwd?: string,
   script: string,
-  scriptArgs: options.Args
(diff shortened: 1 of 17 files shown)


src/commands/run.js
```diff
@@ -19,7 +20,8 @@ export function toRunOptions(
   return {
     cwd: options.string(flags.cwd, 'cwd'),
     script,
-    scriptArgs
+    scriptArgs,
+    scriptFlags: options.toScriptFlags(flags)
```
ajaymathur: should I concat scriptFlags and scriptArgs, and send them as single parameter 🤔

thejameskyle:
I don't think we should be converting meow's output back into an argv string. I think we should pass `argv` through directly.

PR #98, opened 2018-01-07
bolt upgrade

ajaymathur:
bolt upgrade
bolt w upgrade
bolt ws upgrade
🙂


src/commands/__tests__/upgrade.test.js
```diff
@@ -1,4 +1,103 @@
...
+const unsafeYarn: any & typeof yarn = yarn;
+
+async function getDependencyVersion(workspaceDir: string, depName: string) {
+  const pkg = await Package.init(path.join(workspaceDir, 'package.json'));
+  const dirExists = await pathExists(
+    path.join(workspaceDir, 'node_modules', depName)
+  );
+
+  if (!dirExists) {
+    return '1.0.0';
```
ajaymathur: just escape tests where dir is not present

src/commands/workspace/upgrade.js
```diff
@@ -16,5 +18,11 @@ export function toWorkspaceUpgradeOptions(
 }
 
 export async function workspaceUpgrade(opts: WorkspaceUpgradeOptions) {
-  throw new BoltError('Unimplemented command "workspace upgrade"');
+  throw new BoltError(
+    `${chalk.red.bold(
+      '[bolt workspace upgrade]'
+    )} Unable to upgrade dependencies for a single workspace. \nIn order to upgrade a dependency [across all the workspaces] please run ${chalk.green.bold(
```
lukebatchelor: Should we move these to messages.js?

src/commands/workspaces/upgrade.js
```diff
@@ -1,20 +1,44 @@
...
 export async function workspacesUpgrade(opts: WorkspacesUpgradeOptions) {
-  throw new BoltError('Unimplemented command "workspaces upgrade"');
+  let cwd = opts.cwd || process.cwd();
+  let filterOpts = Object.keys(opts.filterOpts);
+
+  if (filterOpts.length) {
+    throw new BoltError(
+      `${chalk.red.bold(
+        '[bolt workspace upgrade]'
+      )} Unable to upgrade dependencies for a single/ some workspace. \nTherefore, flags ${chalk.red.bold(
```
lukebatchelor: And these

src/utils/__tests__/upgradeDependenciesInPackages.test.js
```diff
@@ -0,0 +1,77 @@
...
+    const cwd = await copyFixtureIntoTempDir(__dirname, 'nested-workspaces');
+    const project = await Project.init(cwd);
+
+    await upgradeDependenciesInPackage(project, project.pkg, [
+      { name: 'react', version: '^26.0.0' }
+    ]);
+
+    const deps = (await Project.init(cwd)).pkg.getAllDependencies;
+    const fooDir = getFooDir(project);
+    // expect(await depsAtVersion(project.pkg.dir, { react: '^26.0.0' })).toBe(
```
lukebatchelor: Why's this commented? Not passing?

src/utils/messages.js
```diff
@@ -292,6 +292,12 @@ export function cannotInstallWorkspaceInProject(pkgName: string): Message {
   return `Cannot install workspace "${pkgName}" as a dependency of a project`;
 }
 
+export function cannotUpgradeWorkspaceDependencyInProject(
+  pkgName: string
+): Message {
+  return `Cannot upgrade workspace "${pkgName}" as a dependency of a project`;
```
lukebatchelor: Could we add a tiny bit more info here?

"Cannot upgrade workspace "${pkgName}" as a dependency of a project, workspaces are all symlinked, upgrading would not make sense"

^^ Not the best message either, but I think its better than throwing terms the user might now know.

src/utils/upgradeDependenciesInPackages.js
```diff
@@ -0,0 +1,57 @@
...
+  if (project.pkg.isSamePackage(pkg)) {
+    if (internalDeps.length > 0) {
+      const internalDepNames = internalDeps.map(dep => dep.name);
+      internalDeps.forEach(dep => {
+        logger.error(
+          messages.cannotUpgradeWorkspaceDependencyInProject(dep.name)
+        );
+      });
+      throw new BoltError(
+        'Internal packages are symlinked, there is no need update them'
```
lukebatchelor: Good message, could we move to messages.js too?

lukebatchelor:
Sweet, code all looks good. Lets do some manual testing on monday morn and release

src/commands/workspace/upgrade.js
ajaymathur: I observed logger messages are in messages js. But yeah I should be correct to move to messages js. I will do that.

lukebatchelor:
Sweet, we tested everything manually, all cases working  as expected except for a potential bug in yarn itself.

Running `yarn upgrade` for a package that is not a dependency, but is a transitive dependency, yarn will not only update the yarn.lock, but also add it to the package.json.
Hypothetically this shouldnt be required as the yarn.lock should have all the information it needs to be able to rebuild the node_modules with just the yarn.lock.

We've decided to not code for this behaviour in bolt as it could be considered a bug.

PR #104, opened 2018-01-08
resolving cli flags to commands

ajaymathur:
passing flags to command. Fix : https://github.com/boltpkg/bolt/issues/90

Need Review. 🙂

### DIFF
--- a/src/functions/flagHelper.js
+++ b/src/functions/flagHelper.js
@@ -1,0 +1,47 @@
+// @flow
+
+const IS_BOOLEAN_FLAGS = {
+  '--dev': true,
+  '--peerdev': true
+};
+
+function isFlagWithValue(flag) {
+  return flag.indexOf('=') > -1;
+}
+
+function isFlag(maybeFlag) {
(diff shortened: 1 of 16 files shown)


src/functions/flagHelper.js
```diff
@@ -0,0 +1,47 @@
+// @flow
```
thejameskyle: This shouldn't be in `src/functions`

ajaymathur:
@jamiebuilds @lukebatchelor request review. 🙂

src/utils/flagHelper.js
```diff
@@ -0,0 +1,60 @@
+// @flow
+import * as options from './options';
+
+const BOOLEAN_FLAGS = {
```
lukebatchelor: I thought we discussed this before, that we couldnt have a hard coded list of boolean flags, because it's context dependent?

ajaymathur:
The approach taken here is not going to work:
1. we should not select flags for execution.

PR #111, opened 2018-01-12
Replace jest-fixtures with fixturez

thejameskyle:


luftywiranda13:
@thejameskyle may I know why you replace it? you're also the creator of `jest-fixtures`, right?

thejameskyle:
I redid the API with fixturez cause I like it better

PR #113, opened 2018-01-12
Adding test for options

sendilkumarn:


src/utils/__tests__/options.test.js
```diff
@@ -1,7 +1,76 @@
 // @flow
 import * as options from '../options';
+import {
+  toDependency,
```
sendilkumarn: There is a duplicate method `asDependency` in the project is this needed?

ajaymathur: I believe options is the right place to have this functionality.

ajaymathur: Thanks a lot. but travis didn't trigger on this PR. I will take a loop at that and then merge the PR. 🙂

PR #114, opened 2018-01-13
update readme

chenxsan:
I found that `bolt version` is not implemented yet https://github.com/boltpkg/bolt/blob/e3d0ab41b1b810e158a5cad089c4acf578722451/src/commands/version.js#L15


ajaymathur:
Thanks 🙂

PR #121, opened 2018-01-15
Adding Project add <dependency>

sendilkumarn:
For basic usage, this works really well, like adding the dependency. I have two questions here,

1. What do we have to do when we add the new dependency that is already present in the package? Simply install them? (into the project) I would opt for this given the fact that anybody can go into the project and install the dependencies using `yarn or npm`. 

2. Do we need to keep track of the specific project related dependencies somewhere? In order to make update to be written easier? I would say no to it, since this will create a lot of edge cases.

### DIFF
--- a/src/commands/add.js
+++ b/src/commands/add.js
@@ -6,6 +6,7 @@ import * as logger from '../utils/logger';
 import addDependenciesToPackage from '../utils/addDependenciesToPackages';
 import type { Dependency, configDependencyType } from '../types';
 import { DEPENDENCY_TYPE_FLAGS_MAP } from '../constants';
+import { ProjectAddOptions } from './project/add';
 
 export type AddOptions = {
   cwd?: string,
@@ -42,7 +43,7 @@ export function toAddOptions(
   };
 }
 
-export async function add(opts: AddOptions) {
+export async function add(opts: AddOptions | ProjectAddOptions) {
   let cwd = opts.cwd || process.cwd();
   let project = await Project.init(cwd);
   let pkg = await Package.closest(cwd);
--- a/src/commands/add.js
+++ b/src/commands/add.js
@@ -6,7 +6,7 @@ import * as logger from '../utils/logger';
 import addDependenciesToPackage from '../utils/addDependenciesToPackages';
 import type { Dependency, configDependencyType } from '../types';
 import { DEPENDENCY_TYPE_FLAGS_MAP } from '../constants';
-import { ProjectAddOptions } from './project/add';
+import type { ProjectAddOptions } from './project/add';
 
 export type AddOptions = {
   cwd?: string,
--- a/src/commands/project/__tests__/add.test.js
+++ b/src/commands/project/__tests__/add.test.js
@@ -1,4 +1,23 @@
 // @flow
 import { projectAdd, toProjectAddOptions } from '../add';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as yarn from '../../../utils/yarn';
+import fixtures from 'fixturez';
+import addDependenciesToPackage from '../../../utils/addDependenciesToPackages';
 
-test('bolt project add');
+const f = fixtures(__dirname);
+
+jest.mock('../../../utils/addDependenciesToPackages');
+
+describe('bolt project add', () => {
+  test('adding a project dependency only used by the project', async () => {
+    let tempDir = f.copy('package-with-external-deps-installed');
+
+    await projectAdd(
+      toProjectAddOptions(['project-new-dep'], { cwd: tempDir })
+    );
+
+    expect(addDependenciesToPackage).toHaveBeenCalledTimes(1);
+  });
+});
--- a/src/commands/project/add.js
+++ b/src/commands/project/add.js
@@ -1,16 +1,49 @@
 // @flow
+import addDependenciesToPackage from '../../utils/addDependenciesToPackages';
+import Project from '../../Project';
 import * as options from '../../utils/options';
-import { BoltError } from '../../utils/errors';
+import * as logger from '../../utils/logger';
+import type { Dependency, configDependencyType } from '../../types';
+import { DEPENDENCY_TYPE_FLAGS_MAP } from '../../constants';
 
-export type ProjectAddOptions = {};
+export type ProjectAddOptions = {
+  cwd?: string,
+  deps: Array<Dependency>,
+  type: configDependencyType
+};
 
 export function toProjectAddOptions(
   args: options.Args,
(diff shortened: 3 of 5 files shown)


lukebatchelor:
Heya @sendilkumarn `project add` should be much much simpler than this.
You should be able to completely reuse the add command

```
export  { toAddOptions, type AddOptions } from '../add';
import addDependenciesToPackage from '../../utils/addDependenciesToPackages';
import Project from '../Project';

export async function addToProject(opts: AddOptions) {
  await addDependenciesToPackage(project, project, opts.deps, opts.type);
}
```

sendilkumarn:
With a slight change

> await addDependenciesToPackage(project, project, opts.deps, opts.type); 

 await addDependenciesToPackage(project, project.pkg, opts.deps, opts.type);

lukebatchelor:
Sweet, would you mind adding a single test that just checks that it runs 'add' from the project dir as a sanity check.

sendilkumarn:
Done 👍

src/commands/project/__tests__/add.test.js
```diff
@@ -1,4 +1,22 @@
...
+jest.mock('../../../utils/logger');
+jest.mock('../../../utils/yarn');
+
+describe('bolt project add', () => {
+  test('adding a project dependency only used by the project', async () => {
+    let tempDir = f.copy('package-with-external-deps-installed');
+
+    await projectAdd(toAddOptions(['project-new-dep'], { cwd: tempDir }));
+
+    expect(yarn.add).toHaveBeenCalledTimes(1);
```
lukebatchelor: Sorry, don't want to nitpick, but can this check that it was called with the correct cwd too?

src/cli.js
```diff
@@ -238,9 +238,7 @@ function runCommandFromCli(args: options.Args, flags: options.Flags) {
     return commands.pack(commands.toPackOptions(commandArgs, flags));
   } else if (commandMap.PROJECT[command]) {
     if (commandMap.PROJECT_ADD[subCommand]) {
-      return commands.projectAdd(
-        commands.toProjectAddOptions(subCommandArgs, flags)
-      );
+      return commands.projectAdd(commands.toAddOptions(subCommandArgs, flags));
```
ajaymathur: We should keep using `toProjectAddOptions` , it won't add a lot of code but keep structure simple to follow

src/commands/project/__tests__/add.test.js
sendilkumarn: actually, this is a good catch. There is a problem in completely reusing the add.

ajaymathur: @sendilkumarn Can you please elaborate what is problem coming with this approach?

sendilkumarn: `yarn add` should happen in the pkg directory rather than the `project` or `project.pkg`

We should not set the dependencies. Also, there is no symlink needed here. (since both will be handled by the yarn)

src/cli.js
sendilkumarn: Agree, I was thinking can this be moved to some other util folder or what?

sendilkumarn: Also for testing we can create common folder to stow all the common things that we have.

src/commands/project/__tests__/add.test.js
ajaymathur: I think it should be add it to `project`, what you did is correct. 

for symlink **(since both will be handled by the yarn)**,  I am not sure what so you mean by that?

@lukebatchelor what do you think?

@sendilkumarn 
If you are interested we can continue this discussion on slack - https://join.slack.com/t/boltpkg/shared_invite/enQtMjkzNDk4NjI0OTYyLThhYWRmMDRiMGNiNzEzNWYwYjMwMzJiYjk3ZWE4NGM4N2ZlODA4YzY4NWU5MGM3MDUxNDU5Mzk5ZDg4NTRlODA ?

sendilkumarn: awesome let me join

src/commands/project/add.js
```diff
@@ -1,16 +1,10 @@
...
-): ProjectAddOptions {
-  return {};
-}
-
-export async function projectAdd(opts: ProjectAddOptions) {
-  throw new BoltError('Unimplemented command "project add"');
+export async function projectAdd(opts: AddOptions) {
+  let cwd = opts.cwd || process.cwd();
+  let project = await Project.init(cwd);
+  await addDependenciesToPackage(project, project.pkg, opts.deps, opts.type);
```
lukebatchelor: I believe this is what you're saying above, but to reiterate, this should just be `yarn.add` in the project.
Then the test just needs to check that yarn.add got called with its `cwd` set to projects directory

src/commands/project/__tests__/add.test.js
lukebatchelor: Sorry, might have been confusion here.

We *do* need to use `addDependenciesToPackage` as you have below. 

All `bolt project [cmd]`'s should just be the equivalent to running `cd path/to/project/dir && bolt [cmd]`. The difference here would be the extra checks that `addDependenciesToPackage` does (checking internal/external deps, the side effect of bumping any workspace deps if its changed a version of an external dep, etc).

The test should probably check that `addDependenciesToPackage` is called then, rather than `yarn.add`

sendilkumarn: I completely agree with this

But,

`await yarn.add(project.pkg, externalDepsToInstallForProject, type);` 
The above considers `package` (not project).This method is in fact written on the assumption that wherever we are in the project, do the instructions on the root of the package. 

But `bolt project [cmd]` says don't worry about the `package` do things from the directory where the command is executed.

lukebatchelor: > But bolt project [cmd] says don't worry about the package do things from the directory where the command is executed.

Bolt project cmd, should run in the project dir not in the dir it's executed. See my comment above.

sendilkumarn: Sorry I missed the project word there :)

src/commands/add.js
```diff
@@ -6,6 +6,7 @@ import * as logger from '../utils/logger';
 import addDependenciesToPackage from '../utils/addDependenciesToPackages';
 import type { Dependency, configDependencyType } from '../types';
 import { DEPENDENCY_TYPE_FLAGS_MAP } from '../constants';
+import { ProjectAddOptions } from './project/add';
```
ajaymathur: need type import -> `import type { ProjectAddOptions } from './project/add';`

PR #122, opened 2018-01-16
bolt autoclean

ajaymathur:
- [x] bolt autoclean


src/commands/__tests__/autoclean.test.js
```diff
@@ -0,0 +1,15 @@
...
+
+import { toAutocleanOptions, autoclean } from '../autoclean';
+import * as yarn from '../../utils/yarn';
+
+jest.mock('../../utils/yarn');
+
+const projectDirMock = 'dummyPattern/dummyPath';
+
+describe('bolt autoclean', () => {
+  it('should call yarn link for extenal package', async () => {
```
thejameskyle: wrong description

ajaymathur: and thus I was caught copy/ pasting 😅

PR #124, opened 2018-01-17
Fixes calculation of yarn relative path

lukebatchelor:
Wasn't caught it tests or local testing because I was always running from within the project.
Apologies.


lukebatchelor:
Travis is being useless. Have tested all locally. Gonna merge and release

PR #125, opened 2018-01-18
Implemented login and logout

rajzshkr:
Need clarification:

There is no argument for login and logout. So shall we remove `LoginOptions` and `toLoginOptions`?

### DIFF
--- a/src/commands/__tests__/login.test.js
+++ b/src/commands/__tests__/login.test.js
@@ -1,4 +1,22 @@
 // @flow
 import { login, toLoginOptions } from '../login';
+import * as npm from '../../utils/npm';
 
-test('bolt login');
+jest.mock('../../utils/npm');
+
+describe('bolt login', () => {
+  test('should return the expected cwd', () => {
+    const commandArgs = [];
+    const expectedCwd = 'dummy/path/to/dir';
+    const flags = { cwd: expectedCwd };
+    const result = toLoginOptions(commandArgs, flags);
+    expect(result.cwd).toBe(expectedCwd);
+  });
+
+  test('should call npm login once', async () => {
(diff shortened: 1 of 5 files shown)


src/utils/yarn.js
```diff
@@ -155,3 +155,19 @@ export async function info(cwd: string, spawnArgs: Array<string> = []) {
     tty: true
   });
 }
+
+export async function login(cwd: string) {
```
ajaymathur: can we reuse `cliCommands` function.

ajaymathur:
We should keep using yarn for consistency. 🤔

lukebatchelor:
We use npm for publishing already because of the issues we had with yarn

sendilkumarn:
I do second that we should use `npm`

rajzshkr:
Test case added!!

src/commands/__tests__/login.test.js
```diff
@@ -1,4 +1,30 @@
...
-test('bolt login');
+describe('bolt login', () => {
+  let npmLoginSpy;
+
+  beforeEach(() => {
+    npmLoginSpy = jest.spyOn(npm, 'login');
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
```
ajaymathur: no need to clear mocks.
just mock npm module like
`jest.mock('../../utils/npm')` // just after imports
we are already have config to reset mocks in jest.config.js in root

ajaymathur: sorry it is package json - https://github.com/boltpkg/bolt/blob/master/package.json#L76

rajzshkr: Updated!!

PR #128, opened 2018-01-21
run jest as default script if script test is not present

ajaymathur:


thejameskyle:
This isn't actually what I had in mind for Bolt default scripts.

I wanted a setup more like create-react-app with react-scripts where your package.json is pre-populated with some commands:

```js
{
  "private": true,
  "name": "my-project",
  "scripts": {
    "test": "bolt-scripts test",
    "build": "bolt-scripts build",
    // ...
  },
  "devDependencies": {
    "bolt-scripts": "^1.0.0"
  }
}
```

ajaymathur:
@thejameskyle Thanks, I will give a shot at that. It sounds interesting.

PR #129, opened 2018-01-21
Adding Project Upgrade option

sendilkumarn:


src/commands/project/upgrade.js
```diff
@@ -1,16 +1,58 @@
...
 }
 
 export async function projectUpgrade(opts: ProjectUpgradeOptions) {
-  throw new BoltError('Unimplemented command "project upgrade"');
+  let cwd = opts.cwd || process.cwd();
+  let project = await Project.init(cwd);
+  let pkg = await Package.closest(cwd);
+
+  try {
+    await upgradeDependenciesInPackage(project, pkg, opts.deps, opts.flags);
```
ajaymathur: I think second argument should be `project.pkg` , can you please very if we run this command in workspaces it upgrades in project ?

sendilkumarn: It works. But I have tested with single level of dependency and project.pkg makes sense here

PR #131, opened 2018-01-22
Adding workspace in init

### LINKED ISSUE #119: Adding workspaces automatically when we do bolt init
| Version             | any
| Type                | feature request
| node                | any
| Operating System    | any
| Short Description   | Add workspaces automatically if there are any folders available (with package.json or child directory)
| Detailed description| It would be super awesome if we can add the workspaces automatically when we run `bolt init` for now. Later we can also fetch the child directories package.json and take the common dependencies and add them in here.

sendilkumarn:
closes #119

### DIFF
--- a/src/utils/jsonModifier.js
+++ b/src/utils/jsonModifier.js
@@ -1,0 +1,7 @@
+export default async function addWorkspacesToJson(json: JSONValue) {
+  const workspaces = ['packages/*'];
+  json.bolt = {
+    workspaces
+  };
+  return json;
+}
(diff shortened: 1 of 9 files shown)


src/utils/jsonModifier.js
```diff
@@ -0,0 +1,7 @@
+export default async function addWorkspacesToJson(json: JSONValue) {
```
ajaymathur: need `// @flow` for typecheck to run

src/utils/prompts.js
```diff
@@ -0,0 +1,15 @@
+import inquirer from 'inquirer';
```
ajaymathur: `// @flow`

ajaymathur:
Great stuff 👍 , can you please add test cases too. 🙂

PR #134, opened 2018-01-25
Adding bolt config list command

sendilkumarn:


src/commands/config/list.js
```diff
@@ -1,16 +1,31 @@
...
+};
 
 export function toConfigListOptions(
   args: options.Args,
   flags: options.Flags
 ): ConfigListOptions {
-  return {};
+  if (
+    args.length !== 0 ||
+    Object.keys(DEPENDENCY_TYPE_FLAGS_MAP).length !== 0
```
ajaymathur: what is this for?

src/commands/config/list.js
```diff
@@ -1,16 +1,31 @@
...
+    throw new BoltError('Invalid subcommand, try bolt config list');
+  }
+  return { cwd: options.string(flags.cwd, 'cwd') };
 }
 
 export async function configList(opts: ConfigListOptions) {
-  throw new BoltError('Unimplemented command "config list"');
+  let cwd = opts.cwd || process.cwd();
+  try {
+    await yarn.cliCommand(cwd, 'config list');
```
ajaymathur: this can be `await yarn.cliCommand(cwd, 'config', ['list']);`

as yarn.cliCommand is :

```js
export async function cliCommand(
  cwd: string,
  command: string = '',
  spawnArgs: Array<string> = []
) {
  const localYarn = path.join(await getLocalBinPath(), 'yarn');

  return await processes.spawn(localYarn, [command, ...spawnArgs], {
    cwd,
    tty: true
  });
}
```

src/commands/config/list.js
sendilkumarn: Oops, just found flags are valid.

src/commands/config/list.js
sendilkumarn: makes sense. I will change it.

src/commands/config/list.js
sendilkumarn: wait flags are valid but they are ignored in yarn. So can we also do the same ?

src/commands/config/list.js
```diff
@@ -1,16 +1,28 @@
 // @flow
 import * as options from '../../utils/options';
 import { BoltError } from '../../utils/errors';
+import * as yarn from '../../utils/yarn';
+import { DEPENDENCY_TYPE_FLAGS_MAP } from '../../constants';
```
ajaymathur: nit : this is not used anymore

PR #137, opened 2018-01-26
adding pre-push

sendilkumarn:


ajaymathur:
@sendilkumarn would you like add same info in contributing guide too?

sendilkumarn:
sure.

sendilkumarn:
wait, `prepush` happens automatically right? why do we need to add this to the document?

ajaymathur:
AFAIK `prepush` is not a hook, it might happen if we have a hook (git helper package) installed. I don't think there is any hook in this package. Did this run automatically for you?

sendilkumarn:
yeah

sendilkumarn:
husky

ajaymathur:
Since it is a hook, I think we should not have this ..

sendilkumarn:
I think we can rename it whatsay ?

lukebatchelor:
I'm not sure of the point of this, or why its called "test:complete"?
The pre-commit hook will already run prettier, and CI will already run test and flow validations. If you want to run it before it hits CI, you can, but in the current state its already not possible to push code that isn't prettier'd, flow-checked and tested right? This will just slow down the case where you've already tested everything, so not sure it's useful.

sendilkumarn:
well okayes, let me close this 👍 

But my intention was to have something that does everything. I will run this command before creating a PR / pushing changes

Issue #39, opened 2017-10-22
Fix bolt install validations


thejameskyle:
https://github.com/boltpkg/bolt/pull/36 refactored `bolt install` and moved a bunch of validations into a shared module.

The problem is that these validations previously caused Bolt to exit early, now they aren't run until the linking phase.

These validations should be hoisted up

thejameskyle:
cc @lukebatchelor

lukebatchelor:
Hmmm... I did notice it was moved, do you think itd be better to have the checked up location, or separated into another module?

thejameskyle:
I would break it into three steps:

```js
async function install() {
  let everythingNeededToSymlink = await validateAndReturnInstallSymlinksStuff();
  await yarn.install();
  await symlinkAllTheStuff(everythingNeededToSymlink);
}
```

ajaymathur:
Half way through, I realised we cannot move logic to get symlinks/ and validate for bin before install as it we don't know bin (s) until they are installed..

Should we just have validation logic for external and internal dependencies before install but validation for bin after install?

thejameskyle:
What bin are we validating?

ajaymathur:
Following are the validation for bin 

1. 
```js
pathIsInside(actualBinFile, project.pkg.nodeModules)`
```

2.

```js
if (!actualBinFileRelative) {
      throw new BoltError('${binName} is not a symlink');
}
```

There is a comment on this too:

```js
 // TODO: For now, we'll search through each of the bin files in the Project and find which ones are
 // dependencies we are symlinking. In the future, we should really be going through each dependency
 // and all of its dependencies and checking which ones expose bins so that all the transitive ones
 // are included too
```

PR #144, opened 2018-01-29
[WIP] bolt install validations 

ajaymathur:
related to https://github.com/boltpkg/bolt/issues/39

- [ ] added bolt validate

---

lot..

### DIFF
--- a/src/utils/validateProject.js
+++ b/src/utils/validateProject.js
@@ -16,12 +16,13 @@ export default async function validateProject(project: Project) {
   const { graph: depGraph } = await project.getDependencyGraph(workspaces);
 
   let projectIsValid = true;
+  let invalidMessages = [];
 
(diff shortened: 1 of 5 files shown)


src/utils/validateProject.js
```diff
@@ -41,5 +42,8 @@ export default async function validateProject(project: Project) {
    *     <More Project checks here>
    */
 
-  return projectIsValid;
+  return {
+    invalidMessages,
+    projectIsValid
```
thejameskyle: Don't you need to update `install` now as well?

ajaymathur: yes,, I am planning to call validate command from install to keep single flow for project validation.

thejameskyle: I wouldn't call the command, I would call the util. Calling the command means the validate command is limited by whatever the install command wants to output

PR #149, opened 2018-01-31
Prefix npm_config_user_agent env var set by yarn with bolt version

### LINKED ISSUE #148: Override npm_config_user_agent env var so bolt can be detected
I'd like to use the `npm_config_user_agent` (already set by `npm` and `yarn`) to detect whether `bolt` is being run in my project's `preinstall` script.

Currently npm / yarn set the following values for the `npm_config_user_agent` env var

* yarn (and bolt because it is unchanged): `yarn/1.3.2 npm/? node/v8.9.4 darwin x64`
* npm: `npm/5.6.0 node/v8.9.4 darwin x64`

Ideally bolt will prepend `bolt/<bolt version> ` to the existing env var value set by yarn.

bengummer:
fixes #148

### DIFF
--- a/src/commands/__tests__/install.test.js
+++ b/src/commands/__tests__/install.test.js
@@ -7,6 +7,7 @@ import * as fs from '../../utils/fs';
 import Project from '../../Project';
 import Workspace from '../../Workspace';
 import fixtures from 'fixturez';
+import * as constants from '../../constants';
 
 const f = fixtures(__dirname);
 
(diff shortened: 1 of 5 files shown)


src/commands/__tests__/install.test.js
```diff
@@ -17,6 +18,8 @@ jest.unmock('../install');
 
 const unsafeProcesses: any & typeof processes = processes;
 const unsafeYarn: any & typeof yarn = yarn;
+const unsafeConstants: any & typeof constants = constants;
```
lukebatchelor: Most likely don't need these, we just use them when you need to mock a function and you need to inspect the .mocks property, but still retain other type information.

Just fyi, no need to change

lukebatchelor: Ah, you did do that for processes too

PR #154, opened 2018-02-19
Fix executable linking issues on win32

MarshallOfSound:
The title says it all, basically, we need to read the existing `.cmd` shim first and determine where it is pointing, then create a shim at that target.

### DIFF
--- a/src/utils/fs.js
+++ b/src/utils/fs.js
@@ -49,8 +49,16 @@ function _symlink(src: string, dest: string, type: string) {
   return promisify(cb => fs.symlink(src, dest, type, cb));
 }
 
-function cmdShim(src: string, dest: string) {
-  return promisify(cb => _cmdShim(src, dest, cb));
+function stripExtension(filePath: string) {
+  return filePath.substr(0, filePath.length - path.extname(filePath).length);
+}
+
+async function cmdShim(src: string, dest: string) {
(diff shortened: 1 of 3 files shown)


src/utils/fs.js
```diff
@@ -49,8 +49,16 @@ function _symlink(src: string, dest: string, type: string) {
   return promisify(cb => fs.symlink(src, dest, type, cb));
 }
 
-function cmdShim(src: string, dest: string) {
-  return promisify(cb => _cmdShim(src, dest, cb));
+function stripExtension(filePath: string) {
+  return filePath.substr(0, filePath.length - path.extname(filePath).length);
```
jamiebuilds:
```js
path.basename(filePath, path.extname(filePath));
```

MarshallOfSound: For input `/project/test.cmd` wouldn't that give just `test`, the shim cmd function requires an absolute path (`/project/test`).

I initially had `path.join(path.dirname(filePath), path.basename(filePath, path.extname(filePath))` but thought that for speed it would be better just to substring the extension off the end. Thoughts?

jamiebuilds: I'd rather use built-in functions than hand optimising something so small

MarshallOfSound: @jamiebuilds Done 👍

PR #158, opened 2018-03-02
[WIP] Sample approach to handle commandMap

ajaymathur:
If this approach looks ok. I make change the whole repo. It will not be a big change in command files.


src/commands/index.js
```diff
@@ -1,48 +1,6 @@
 // @flow
 module.exports = {
   ...require('./add'),
```
ajaymathur: ignore this file change, it is just to that we can test stuff, everything will come back in once all commands are updated.

src/commands/workspace/index.js
```diff
@@ -1,10 +1,22 @@
 // @flow
```
ajaymathur: all subcommand file will need to be updated 🙂

src/cli.js
```diff
@@ -9,398 +9,19 @@ import cleanStack from 'clean-stack';
...
-  } else if (commandMap.UNLINK[command]) {
-    return commands.unlink(commands.toUnlinkOptions(commandArgs, flags));
-  } else if (commandMap.UPGRADE[command]) {
-    return commands.upgrade(commands.toUpgradeOptions(commandArgs, flags));
-  } else if (commandMap.UPGRADE_INTERACTIVE[command]) {
-    return commands.upgradeInteractive(
-      commands.toUpgradeInteractiveOptions(commandArgs, flags)
+  if (!command) return await commands['install'](flags, commandArgs);
+
+  if (command && commands[command])
```
ajaymathur: we will get command as:
```
{ 
  add: [add Function]
}
```

So we can just call is as commands['add']  -- given command is add ( bolt add ...  )

src/commands/add.js
```diff
@@ -43,7 +43,8 @@ export function toAddOptions(
   };
 }
 
-export async function add(opts: AddOptions | ProjectAddOptions) {
+export async function add(flags: options.Flags, commandArgs: Array<string>) {
```
ajaymathur: related to https://github.com/boltpkg/bolt/pull/158/files#r171766571

src/cli.js
ajaymathur: we are passing all `flags, commandArgs, subCommand, subCommandArgs` but can ignore in commands?

ajaymathur:
We will not implement this change at this time. 🙂

PR #165, opened 2018-04-18
Fixes bolt install to use local yarn instead of global

lukebatchelor:
Turns our we never actually made bolt install use the local one at all.

Change was quite simple, tests were a pain to refactor, but I think they are a lot cleaner now.

I also fixed the issue with the long paths being shows in the logs by adding a 'useBasename' flag to spawn and turning that on for `yarn install`, and all `yarn run` commands.

I couldnt properly test that locally for some reason, `bolt install` for example wouldnt show the command in the front of the logs. But it should be safe to test in CI either way.

### DIFF
--- a/src/utils/__tests__/yarn.test.js
+++ b/src/utils/__tests__/yarn.test.js
@@ -4,16 +4,20 @@ import * as path from 'path';
 import * as yarn from '../yarn';
 import * as processes from '../processes';
 import Project from '../../Project';
+import * as constants from '../../constants';
 import fixtures from 'fixturez';
+import containDeep from 'jest-expect-contain-deep';
 
 const f = fixtures(__dirname);
 
 jest.mock('../processes');
 
-async function getLocalBinPath(): Promise<string> {
-  return await projectBinPath();
+async function getLocalYarnPath(): Promise<string> {
+  const boltBinDir = await projectBinPath();
+  return path.join(boltBinDir, 'yarn');
 }
 const unsafeProcesses: any & typeof processes = processes;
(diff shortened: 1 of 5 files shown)


src/utils/__tests__/yarn.test.js
```diff
@@ -24,7 +28,81 @@ function assertSpawnCalls(expectedProcess, expectedArgs, expectedCwd) {
   expect(spawnCalls[0][2].cwd).toEqual(expectedCwd);
 }
 
+let yarnUserAgent = 'yarn/7.7.7 npm/? node/v8.9.4 darwin x64';
+
 describe('utils/yarn', () => {
+  describe('install()', () => {
```
lukebatchelor: I've written these tests to be more self contained and not do all that funky nested beforeEach business ;P

`jest-expect-contain-deep` ended up being super useful btw, cheers

src/utils/yarn.js
```diff
@@ -22,6 +22,23 @@ function depTypeToFlag(depType) {
   return flag ? `--${flag}` : flag;
 }
 
+export async function install(cwd: string, pureLockfile?: boolean) {
```
lukebatchelor: This was the only thing I debated, the signature here is a bit weird.

```
install(cwd, pureLockfile)
```

I feel like maybe it should just take a single opts object maybe? It's only internal API so I'm not super worried about it, and am happy to change if you have a preference

lukebatchelor:
I'm pretty confident this will all be fine, so going to merge and release to give it a proper test. 
Do give a yell if you want me to change anything though.

PR #168, opened 2018-04-19
Adds checks for config name and version fields to project validation

lukebatchelor:


src/utils/validateProject.js
```diff
@@ -27,6 +27,23 @@ export default async function validateProject(project: Project) {
     }
   }
 
+  // Workspaces always have a name and a version in their configs
+  for (let pkg of packages) {
```
lukebatchelor: Turns out the error we hit was a tiny bit of an edge case, in that *most* packages would actually end up throwing errors before this ever gets run. The only reason we hit the error is because the package without a version is not a dependency of any other package in the repo.

For that same reason the "name" check here is mostly redundant too, whilst we build the dependency graph, you'll already get an error if any package is missing a name.

I've left it here for semantic correctness though. It's best not to make code dependent on side effects that happen somewhere else, and this is not an expensive task

PR #174, opened 2018-05-23
Fixes logging and makes much clearer what is happening during publish

lukebatchelor:


lukebatchelor:
Basically just tried to make the code easier to read, removed the incorrect logging that happened when our version is equal to the published version and added an extra log explaining when we choose that a package is being published.

It creates some repeated information, but it would have helped immensely when debugging today.

src/commands/publish.js
```diff
@@ -25,27 +25,6 @@ export function toPublishOptions(
...
 
 async function getUnpublishedPackages(packages) {
-  let semverGtCheckFailWithWarning = (
-    pkgLocalVersion: string,
-    pkgPublishedVersion: string,
-    pkgName: string
-  ): boolean => {
-    let shouldPublish = semver.gt(pkgLocalVersion, pkgPublishedVersion);
-
-    // show a warning message if package is not published since pkgPublishedVersion > pkgLocalVersion
```
lukebatchelor: So mistake here is that the `shouldPublish` check above is correct, but not all cases where we arent publishing are errors (when we are equal for example).

src/commands/publish.js
```diff
@@ -55,22 +34,31 @@ async function getUnpublishedPackages(packages) {
         name: config.getName(),
         localVersion: config.getVersion(),
         isPublished: response.published,
-        newVersion: response.pkgInfo.version || ''
+        publishedVersion: response.pkgInfo.version || ''
```
lukebatchelor: Just a clearer name

src/utils/messages.js
```diff
@@ -271,7 +271,22 @@ export function noUnpublishedPackagesToPublish(): Message {
   return 'No unpublished packages to publish';
 }
 
-export function notPublishingPackage(
```
lukebatchelor: Slightly renamed here to be different from `willPublishPackage` above

PR #178, opened 2018-06-29
Accept arrays for filter options

### LINKED ISSUE #172: ignore/only{Fs} should accept arrays
```js
getWorkspaces({
  cwd,
  ignoreFs: [
    'packages/css-packs/*',
    'build/*',
    'website'
  ]
});
```

https://github.com/boltpkg/bolt/blob/b2f52a25311df2152794002f1d8792fadebe8360/src/functions/getWorkspaces.js#L7-L10

via @jamiebuilds

sidoshi:
Fixes #172

### DIFF
--- a/src/__tests__/Project.test.js
+++ b/src/__tests__/Project.test.js
@@ -155,76 +155,111 @@ describe('Project', () => {
   test('filterPackages() with only flag', async () => {
     let project = await Project.init(f.find('nested-workspaces'));
     let packages = await project.getPackages();
-    let filtered = await project.filterPackages(packages, {
+    let filteredString = await project.filterPackages(packages, {
+      only: 'foo'
+    });
+    let filteredArray = await project.filterPackages(packages, {
       only: ['foo', 'bar']
     });
-    assertPackages(filtered, ['bar', 'foo']);
+    assertPackages(filteredString, ['foo']);
+    assertPackages(filteredArray, ['bar', 'foo']);
   });
 
   test('filterPackages() with only flag as glob', async () => {
     let project = await Project.init(f.find('nested-workspaces'));
     let packages = await project.getPackages();
-    let filtered = await project.filterPackages(packages, {
+    let filteredString = await project.filterPackages(packages, {
+      only: 'ba*'
+    });
+    let filteredArray = await project.filterPackages(packages, {
       only: ['ba*']
     });
-    assertPackages(filtered, ['bar', 'baz']);
+    assertPackages(filteredArray, ['bar', 'baz']);
+    assertPackages(filteredString, ['bar', 'baz']);
   });
 
   test('filterPackages() with ignore flag', async () => {
     let project = await Project.init(f.find('nested-workspaces'));
     let packages = await project.getPackages();
-    let filtered = await project.filterPackages(packages, {
+    let filteredString = await project.filterPackages(packages, {
+      ignore: 'bar'
+    });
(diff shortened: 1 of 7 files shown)


src/utils/options.js
```diff
@@ -35,10 +35,14 @@ export function number(val: mixed, name: string): number | void {
...
-  if (flags.ignore) filterOpts.ignore = string(flags.ignore, 'ignore');
-  if (flags.ignoreFs) filterOpts.ignoreFs = string(flags.ignoreFs, 'ignoreFs');
+  let split = (s: string | void): Array<string> =>
+    typeof s === 'string' ? s.split(',').filter(Boolean) : [];
+
+  if (flags.only) filterOpts.only = split(string(flags.only, 'only'));
+  if (flags.onlyFs) filterOpts.onlyFs = split(string(flags.onlyFs, 'onlyFs'));
+  if (flags.ignore) filterOpts.ignore = split(string(flags.ignore, 'ignore'));
+  if (flags.ignoreFs)
+    filterOpts.ignoreFs = split(string(flags.ignoreFs, 'ignoreFs'));
```
ajaymathur: the string function check if `typeof val !== string`, is should fail in case of arrays?

sidoshi:
I don't think we will be receiving array here as these  options are received from the client args

ajaymathur:
Thanks for clarifying, LTGM 🎉

src/__tests__/Project.test.js
```diff
@@ -156,16 +156,16 @@ describe('Project', () => {
     let project = await Project.init(f.find('nested-workspaces'));
     let packages = await project.getPackages();
     let filtered = await project.filterPackages(packages, {
-      only: 'foo'
+      only: ['foo', 'bar']
```
lukebatchelor: In updating these tests you're also changing the code paths they used to test. 

Would it be possible to keep them as close to the old tests as possible and add new tests for the new functionality?

i.e

```
let filtered = await project.filterPackages(packages, {
-  only: 'foo'
+  only: ['foo']
```

That way we know the base cases still work?

lukebatchelor:
Hiya, thanks for the contribution.

My only thought is that by not accepting strings as well we need to make this a breaking change. I think keeping it backwards compatible would be preferred.

src/utils/__tests__/options.test.js
```diff
@@ -29,11 +29,17 @@ describe('options', () => {
   });
 
   test('toFilerOptions', () => {
-    expect(toFilterOpts({ only: 'only' })).toEqual({ only: 'only' });
-    expect(toFilterOpts({ onlyFs: 'onlyFs' })).toEqual({ onlyFs: 'onlyFs' });
-    expect(toFilterOpts({ ignore: 'ignore' })).toEqual({ ignore: 'ignore' });
-    expect(toFilterOpts({ ignoreFs: 'ignoreFs' })).toEqual({
-      ignoreFs: 'ignoreFs'
+    expect(toFilterOpts({ only: 'onlyOne,onlyTwo' })).toEqual({
```
lukebatchelor: Same here, could these be extra tests instead of replacing the old ones?

sidoshi:
Yes, it would be better to keep this backward compatible. 

I thought it would only be a breaking change internally as there would be no breaking change for users using the CLI. But I didn't realize we expose that function from library API as well. 

I will work on making this backward compatible and also updating the tests over the weekend.

sidoshi:
I have added string support to make this change backwards compatible

Issue #181, opened 2018-07-24
Bolt publish errors don't draw enough attention to the useful error message


lukebatchelor:
Example error: https://bitbucket.org/atlassian/atlaskit-mk-2/addon/pipelines/home#!/results/25485

If you look at the error messages there we get a bunch of colorful (completely correct) error messages:

```
error Package "@atlaskit/adf-utils" must depend on the current version of "@atlaskit/editor-common": "12.0.0" vs "^11.4.0"
error Package "@atlaskit/conversation" must depend on the current version of "@atlaskit/editor-common": "12.0.0" vs "^11.4.6"
error Package "@atlaskit/editor-bitbucket-transformer" must depend on the current version of "@atlaskit/editor-common": "12.0.0" vs "^11.3.8"
error Package "@atlaskit/editor-confluence-transformer" must depend on the current version of "@atlaskit/editor-common": "12.0.0" vs "^11.3.8"
```

The useful error message is the one down the bottom:

```
Error: @atlaskit/adf-utils has a dependency on @atlaskit/editor-common at ^11.4.0, however the new version of 12.0.0 leaves this range. You will need to make a new changeset that includes an update to @atlaskit/adf-utils
    at Object.updatePackageVersions (/opt/atlassian/pipelines/agent/build/node_modules/bolt/dist/modern/functions/updatePackageVersions.js:92:15)
    at <anonymous>
```

It would be nice to make this message much, much more prominent. Ie using color, a section break or a box around it

| Title               | Description
| --------------------|------------
| Version             | 0.20.7

PR #182, opened 2018-07-24
Updates an error message to be much more prominent

lukebatchelor:
It now looks something like this:
alt text

I realised whilst looking at this that we probably shouldnt have that error message in bolt itself because it assumes the existance of changesets. Im happy to leave for now though until `bolt` has a blessed way of handling this (it'll be a minor text change).

### DIFF
--- a/src/utils/messages.js
+++ b/src/utils/messages.js
@@ -30,6 +30,10 @@ function cmd(str) {
   return chalk.bgBlack.magenta(`\`${str}\``);
 }
 
+const importantSeparator = chalk.red(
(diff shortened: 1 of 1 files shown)


lukebatchelor:
Super small change so I'll merge tomorrow if you don't mind @jamiebuilds

lukebatchelor:
@Noviny FYI, see above for fix to the issue you talked about, but also note that we probably want to change the messaging at some point

lukebatchelor:
Fixes #181

PR #188, opened 2018-08-23
Separate publish CLI from exposed publishPackages function

lachlanhunt:


lukebatchelor:
Awesome looks great.
Note for anyone that's looking at this later, the issue here was that we expose `functions` and `commands` from bolt. Functions being the ones that are simplified and created for external consumption, commands being a way to call the same code that gets called when you run a bolt command (`bolt publish` for example).

We were a bit lazy with the `bolt publish` command and essentially used it as a command that could be consumed (main difference being that `functions` return values, commands should exit with status codes).

This cleans up that abstraction so that `bolt publish` returns correctly, but that `bolt.publishPackages()` can be called and you'll get a list of published and unpublished packages.

**Note** this technically makes it a breaking change as `bolt.publish()` will no longer return a list of packages. Consumers will need to swap to `bolt.publishPackages()`

Issue #189, opened 2018-08-27
Bolt publish unable to authenticate using npm credentials.


lukebatchelor:
We've actually seen this issue a few times before, but have a much better workaround now.

**The problem**: When running `bolt publish`, things will work as expected. When running `bolt publish` from within a script, you are unable to authenticate. This includes using `bolt.publish({})` from a node script (if it is called using `bolt`).

The issue turned out to be the same as what was reported[here](https://github.com/yarnpkg/yarn/issues/2935#issuecomment-355292633).

> ...seems like yarn is setting env variable which is npm using for publishing npm_config_registry this feels like yarn issue to me.

This was confirmed to be the case here as well.

There are a few workarounds possible for others with this issue:

* Simply not using yarn or bolt to run the script. `npm` and `npx` will work fine.
* Adding a `.yarnrc` to the project with `registry "https://registry.npmjs.org/"` in it.
  * Tested, working here.
* Using the `--registry` flag when publishing
  * Should work, but testing has not been successful so far
* Manually overriding the `npm_config_registry` variable when running `npm publish` in bolt
  * Not tested yet, but would be useful to add into `bolt`.

We are using option 2 in a couple of places atm, but would like to implement option 4 in bolt itself if possible.

lukebatchelor:
Fixed in #190

PR #190, opened 2018-09-04
Fixes bug when running publish using bolt or yarn

lukebatchelor:


lukebatchelor:
Fixes #189

jamiebuilds:
Err... This causes the registry to _always_ be set to `https://registry.npmjs.org/` so if you are trying to publish to a private internal registry, you can't.

Issue #126, opened 2018-01-18
Binary exported by workspace package causes "bolt install" failure


bengummer:
| Title               | Description
| --------------------|------------
| Version             | 0.19.2
| Type                | Issue
| node                | 8.9.4
| Operating System    | macOS 10.13.2
| Short Description   | Binary exported by workspace package causes "bolt install" failure
| Detailed description| If I have a bolt workspace containing a package that has a `"bin"` entry in its `package.json` file, and also have another package in the same workspace that depends on the first bin-exporting package, then running `bolt` results in an error.

The error displayed is:

```sh
$ bolt
⚡️   bolt v0.19.2 (node v8.9.4)
🔎   [1/3] Validating project...
📦   [2/3] Installing project dependencies...
yarn install v1.3.2
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 📃  Building fresh packages...
✨  Done in 0.14s.
🔗   [3/3] Linking workspace dependencies...
error my-bin is linked to a location outside of project node_modules: ../../packages/my-bin/index.js
```

Demo repo here: https://github.com/bengummer/bolt-binary-issue-demo

`yarn install` works in the demo repo, `bolt install` doesn't

lukebatchelor:
Ah. Thats interesting. We probably need to either symlink workspace's bins to the project bin, then dependents symlink to that, or modify this check to not happen with internal deps. 

Probably option 2

d4rkr00t:
Hm, it works fine here: https://bitbucket.org/atlassian/atlaskit-mk-2/src/394ebda17e5de5bd3eeb39d0574380fe283a9248/build/webpack-config/package.json#package.json-4 and https://bitbucket.org/atlassian/atlaskit-mk-2/src/394ebda17e5de5bd3eeb39d0574380fe283a9248/website/package.json#package.json-58

bengummer:
Yeah the following tests looks like they test internal dependencies so i'm not sure why my demo repo has the issue.

https://github.com/boltpkg/bolt/blob/41e1ed61b2/src/utils/__tests__/symlinkPackageDependencies.test.js#L104

jamiehodge:
I ran into this issue as well

jamiebuilds:
Moving to #157

Issue #187, opened 2018-08-22
Bolt does not wire up workspace packages' clis


padmaia:
| Title               | Bolt does not wire up workspace packages' clis
| --------------------|------------
| Version             | 0.20.7
| Type                | Issue
| node                | 8.11.3
| Operating System    | macOS 10.13.6
| Short Description   | Bolt does not seem to wire up clis exposed by workspace packages. See https://github.com/padmaia/bolt-package-with-cli-broken for a minimal reproducible example.

lukebatchelor:
Ha. Damn. We literally ran into this just now...
I might try to take a look on the weekend or next week if you don't get to it first.

We'll work around it for now

lukebatchelor:
Oh, I misunderstood this issue again, this *was* fixed in #191 .

The issue that's not fixed is adding workspace binaries to the Projects binary too.
Actually, I need to double check whether it should. Either way, workspaces that have dependencies on each other will have binaries symlinked now (bolt 0.21.1).

https://github.com/boltpkg/bolt/blob/593937a6055f89d707e2321accb438e2adbf3b33/src/commands/__tests__/install.test.js#L143

lukebatchelor:
So I completely misunderstood this a few times. It was *definitely* fixed in #194 and released in bolt 0.22.2 for anyone looking

PR #157, opened 2018-02-22
Fix issue with cross-dep bins not symlinking

jamiebuilds:
Fixes #126

PR #191, opened 2018-09-06
Fix issue with cross-dep bins not symlinking

lukebatchelor:
This is simply finishing up #157 . It was a bit out of date.
I thought it would also fix  _ #187 _ but this just solves the cross linking problem, it doesn't make them accessible at the root.


lukebatchelor:
Fixes #126

PR #192, opened 2018-09-14
add --serial and --parallel options

jamiebuilds:
A whole bunch of people have asked for these. 

**Runs as parallel as possible while respecting the dep graph:** (safest)

```sh
bolt workspaces exec -- cmd
```

**Runs as parallel as possible ignoring the dep graph:** (fastest)

```sh
bolt workspaces exec --parallel -- cmd
```

**Runs sequentially:** (slowest)

```sh
bolt workspaces exec --serial -- cmd
```

I just added this into `runPackageTasks` so anything that runs across all your packages should now support these flags.

I also want to add a `--concurrency <n>` option for setting a limit of max processes to spawn at any given time. I've set up the basis for adding that, but I wasn't sure how I wanted to do it in code yet so I left it out as a TODO.

cc @kayneb

### DIFF
--- a/src/Project.js
+++ b/src/Project.js
@@ -4,7 +4,7 @@ import includes from 'array-includes';
 import semver from 'semver';
 import Package from './Package';
 import Config from './Config';
-import type { FilterOpts } from './types';
+import type { SpawnOpts, FilterOpts } from './types';
 import * as fs from './utils/fs';
 import * as logger from './utils/logger';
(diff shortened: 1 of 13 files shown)


lukebatchelor:
Sweet, LGTM.
Will release a minor version?

lukebatchelor:
Released in bolt 0.22.0

PR #194, opened 2018-09-17
Adds symlinking of package binaries to the project at install

### LINKED ISSUE #187: Bolt does not wire up workspace packages' clis

lukebatchelor:
Fixes #187 

* Adds symlinking of package binaries to the project at install time.
* Removes check that all project bin symlinks must point to project node_modules
* Fixes some publish tests

### DIFF
--- a/src/commands/__tests__/publish.test.js
+++ b/src/commands/__tests__/publish.test.js
@@ -3,7 +3,6 @@ import { publish, toPublishOptions } from '../publish';
 import fixtures from 'fixturez';
 import * as logger from '../../utils/logger';
 import * as publishUtils from '../../utils/publish';
-import { BoltError } from '../../utils/errors';
 
 const f = fixtures(__dirname);
 
@@ -44,7 +43,7 @@ describe('bolt publish', () => {
     expect(unsafePublishUtil).toHaveBeenCalledTimes(1);
   });
 
-  test('should throw if publishing failed for 1 or more packages', async () => {
+  test('should throw if publishing failed > 1 time', async () => {
     let cwd = f.find('simple-project');
     unsafePublishUtil.mockReturnValue([
       {
@@ -59,12 +58,11 @@ describe('bolt publish', () => {
       }
     ]);
 
-    expect(publish({ cwd })).rejects.toBeInstanceOf(BoltError);
+    await expect(publish({ cwd })).rejects.toBeInstanceOf(Error);
     expect(unsafePublishUtil).toHaveBeenCalledTimes(1);
   });
 
-  xtest('should log successful and unsuccessful packages', async () => {
-    console.log('TEST STARTS HERE');
+  test('should log successful and unsuccessful packages', async () => {
     let cwd = f.find('simple-project');
     unsafePublishUtil.mockReturnValue([
       {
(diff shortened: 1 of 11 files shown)


lukebatchelor:
@jamiebuilds Just wanted to get your thoughts on this.

We currently have this check in the symlinking step.

https://github.com/boltpkg/bolt/blob/838010d48ae0ed87be067d391b56881e4dfd48bf/src/utils/symlinkPackageDependencies.js#L130-L134

This causes this feature to fail if you run `install` twice (because there will be a symlink to somewhere outside of node_modules).

In your opinion, should we remove this check (not 100% sure if it's required?) or should we just loosen it to also check if the symlink points to a `Package` (or even stricter, a `Packages` binary).

Thoughts

src/commands/__tests__/publish.test.js
```diff
@@ -94,7 +92,7 @@ describe('bolt publish', () => {
       }
     ]);
 
-    expect(publish({ cwd })).rejects.toBeInstanceOf(BoltError);
+    await expect(publish({ cwd })).rejects.toBeInstanceOf(Error);
```
lukebatchelor: @lachlanhunt Just FYI, found why the test below was failing and also found that this test shouldn't have been passing. :S

We weren't awaiting on the expect.

There's another interesting thing here, we can't check for instanceOf(BoltError) because on lower versions of node it just comes through as "Error". Weird.

src/utils/symlinkPackageDependencies.js
```diff
@@ -127,12 +127,6 @@ export default async function symlinkPackageDependencies(
       actualBinFileRelative
     );
 
-    if (!pathIsInside(actualBinFile, project.pkg.nodeModules)) {
```
lukebatchelor: Discussed with @jamiebuilds and we're removing this as it was for future plans around linking multiple bolt repos that didn't come to fruition.

src/utils/symlinkPackageDependencies.js
```diff
@@ -191,7 +185,10 @@ export default async function symlinkPackageDependencies(
       let src = path.join(depWorkspace.pkg.dir, String(binPath));
       let dest = path.join(pkg.nodeModulesBin, binName);
 
-      symlinksToCreate.push({ src, dest, type: 'exec' });
+      // Just in case the symlink is already added (it might have already existed in the projects bin/)
+      if (!symlinksToCreate.find(symlink => symlink.dest === dest)) {
```
lukebatchelor: Just to explain this check:

I noticed that when running `install` twice in a row I'd get an error about a symlink already being created. Originally I assumed this must have been from my added code (which hypothetically should be fine, the symlink command rimrafs existing symlinks if they are there).

Upon investigation I found it it was happening in `symlinkPackageDependencies` because a symlink was being added to `symlinksToCreate` twice, once because it was in the root dependencies and once because it was an internal dependency.

This check was the least intrusive way to fix the issue.

A couple of other solutions could be: 
* Remove all the code that symlinks binaries between packages and let this be taken care of by `symlinkPackagesBinariesToProject`. It would copy all binaries from packages into the root .bin and then we'd automatically pick them all up anyway. There's no problem with this imo, the main difference would be that all packages would now have bin's from other internal packages, where as previously they would only get them if they had a dependency (this is not much different to all packages getting a link to all root external bins imo)
* Run the symlinking code in serial rather than parallel. Not a great solution, it would solve the problem because each time we tried to make a symlink it would always delete first if there was an existing one, this would be fairly slow however.
* We could rethink how we do all the symlinking of binaries into packages. This currently actually takes a very long time for some reason. We could investigate possibly symlinking the bin directory instead? This would be the most dangerous change imo.

PR #214, opened 2019-01-04
Fixes #90 by passing flags down as args for run commands

### LINKED ISSUE #90: pass on the -- flags to scripts

lukebatchelor:
Thanks heaps for the patience everyone!

The solution I ended up going with is basically what was outlined [here](https://github.com/boltpkg/bolt/issues/90#issuecomment-451311413)

With each of these ways to run a script

```
bolt foo
bolt run foo
bolt p run foo
bolt w @dovetail/bar run foo
bolt ws --only="@dovetail/bar" run foo
bolt ws exec --only="@dovetail/bar" -- yarn foo
```

The way you would now pass flags is exactly as you'd expect

```
bolt foo --flag1 --flag2=5
bolt run foo --flag1 --flag2=5
bolt p run foo --flag1 --flag2=5
bolt w @dovetail/bar run foo --flag1 --flag2=5
bolt ws exec --only="@dovetail/bar" -- yarn foo  --flag1 --flag2=5
```

The only different one is the `ws run` command

```
bolt ws --only="@dovetail/bar" run foo -- --flag1 --flag2=5
```

The things we are trading off here are

* Complexity of code (simpler to do it this way)
* Complexity for the consumer (slightly more complicated this way)
* Ambiguity of commands (this ensures there can be no ambiguity)

**The implementation**

This wasn't as clean as I'd hoped. You'll see the comments I added, but basically, because meow has already parsed all our flags, there's no perfect way to pass them down accurately without reaching out and using `process.argv` directly.

This also makes writing tests for this a much bigger PITA. I'm willing to come back to those at some point after I do some investigations today.

Leaving this PR up now to get some feedback.

CC: @jamiebuilds @marcins @d4rkr00t @bradleyayers @drew-walker

### DIFF
--- a/src/commands/project/run.js
+++ b/src/commands/project/run.js
@@ -23,17 +23,14 @@ export function toProjectRunOptions(
    * and args on the other side e.g `bolt run test src/* -- --watch --bail` which is further away from
    * how yarn handles things and is more complicated for the consumer.
    */
-  let [script, ...scriptArgs] = args;
-  // the flags that we'll be passing in as args
-  let flagArgs = process.argv
-    .slice(3)
-    .filter(
-      (arg, idx) => args.indexOf(arg) === -1 && !(arg === 'run' && idx === 0)
-    );
+  let [script] = args;
+  const scriptArgIdx = process.argv.indexOf(script);
+  // the flags that we'll be passing in as args start after the script name, and we'll pass them all directly
+  const scriptArgs = process.argv.slice(scriptArgIdx + 1);
   return {
     cwd: options.string(flags.cwd, 'cwd'),
     script,
-    scriptArgs: [...scriptArgs, ...flagArgs]
+    scriptArgs
   };
 }
 
--- a/src/commands/project/run.js
+++ b/src/commands/project/run.js
@@ -15,11 +15,25 @@ export function toProjectRunOptions(
   args: options.Args,
   flags: options.Flags
 ): ProjectRunOptions {
+  /**
+   * Unfortunately, in run commands, we are unable to use meow's parsed flags and args as we really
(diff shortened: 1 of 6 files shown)


src/commands/project/run.js
```diff
@@ -15,11 +15,25 @@ export function toProjectRunOptions(
...
+   * and args on the other side e.g `bolt run test src/* -- --watch --bail` which is further away from
+   * how yarn handles things and is more complicated for the consumer.
+   */
   let [script, ...scriptArgs] = args;
+  // the flags that we'll be passing in as args
+  let flagArgs = process.argv
+    .slice(3)
+    .filter(
+      (arg, idx) => args.indexOf(arg) === -1 && !(arg === 'run' && idx === 0)
+    );
```
jamiebuilds: This should probably be inside a util

lukebatchelor: As discussed, might just try removing the scriptArgs completely and removing the extra indexOf here.
At that point, its quite a simple function, still want it in a util?

src/commands/project/run.js
```diff
@@ -15,11 +15,25 @@ export function toProjectRunOptions(
...
   return {
     cwd: options.string(flags.cwd, 'cwd'),
     script,
-    scriptArgs
+    scriptArgs: [...scriptArgs, ...flagArgs]
```
willmendesneto: 👏

lukebatchelor:
@jamiebuilds changed those scripts to just use argv now, much cleaner. Going to merge.

lukebatchelor:
Hey @jamiebuilds I've skipped the tests for now and I'm leaning towards just changing how we pass args. I'll leave this up for a couple more days in case you've got thoughts 👌

Issue #207, opened 2018-11-30
Error during install with command line packages on Windows


kireerik:
| Title               | Description
| --------------------|------------
| Version             | 0.22.4
| Type                | Issue
| node                | 11.3.0
| Operating System    | Windows 10

I am recieving the following error on `bolt install`:
```bash
[4/4] Linking workspace binaries...
error Error: Can't read shim path from 'project/packages/project/index.js', it doesn't appear to be a cmd-shim
error     at cb (AppData/Local/Yarn/Data/global/node_modules/bolt/dist/modern/utils/fs.js:154:75)
error     at AnyPromise (AppData/Local/Yarn/Data/global/node_modules/typeable-promisify/index.js:13:3)
error     at new Promise (<anonymous>)
error     at promisify (AppData/Local/Yarn/Data/global/node_modules/typeable-promisify/index.js:12:40)
error     at readCmdShim (AppData/Local/Yarn/Data/global/node_modules/bolt/dist/modern/utils/fs.js:154:42)
error     at cmdShim (AppData/Local/Yarn/Data/global/node_modules/bolt/dist/modern/utils/fs.js:96:88)
error     at createWindowsSymlink (AppData/Local/Yarn/Data/global/node_modules/bolt/dist/modern/utils/fs.js:122:18)
error     at Object.symlink (AppData/Local/Yarn/Data/global/node_modules/bolt/dist/modern/utils/fs.js:134:18)
```

ajaymathur:
Hey @kireerik , can also push a minimal reproducible example repository? It will be easier for me to debug and understand the issue. 🙂

kireerik:
Hi @ajaymathur, I think you can reproduce this issue in this repository: https://github.com/kireerik/refo

You will get the error for this package first (https://github.com/kireerik/refo/tree/master/index/index/core/index) and if you remove these lines from it's package.json, you will no longer get the error for that package:
https://github.com/kireerik/refo/blob/5f5757cc1110e55abc851116e8b55cbc6ccfbe78/index/index/core/index/package.json#L4-L6

After this, you will get the error for this package (https://github.com/kireerik/refo/tree/master/index/index/index). This package it self is very minimal. Also if you remove these lines here (https://github.com/kireerik/refo/blob/5f5757cc1110e55abc851116e8b55cbc6ccfbe78/index/index/index/package.json#L4-L6) then you will no longer get the error as well.

[...]

PR #215, opened 2019-01-06
Proposal fix for Issue #207

Bloodb0ne:
Recently i have encountered the same issues as #207 . It looks like packages in the workspaces referencing another package in the same workspace dont have a shim create thats why `readCmdShim` fails, because it expects a shim file. 
My solution just uses the `src` variable to create the new shim file after that if there occurs another reference to that package it will read the shim file and use that path.

### DIFF
--- a/src/utils/fs.js
+++ b/src/utils/fs.js
@@ -57,10 +57,15 @@ function stripExtension(filePath: string) {
 }
 
 async function cmdShim(src: string, dest: string) {
-  const currentShimTarget = path.resolve(
-    path.dirname(src),
-    await readCmdShim(src)
-  );
+  var currentShimTarget = src;
+  try{
+    var currentShimTarget = path.resolve(
+      path.dirname(src),
+      await readCmdShim(src)
+    );
+  }catch(err){
+    
+  }
   await promisify(cb => _cmdShim(currentShimTarget, stripExtension(dest), cb));
 }
 


lukebatchelor:
Hiya. Thanks for the contribution. 

None of the core maintainers do much of the way of windows development so finding and fixing these sorts of issues can be a pain for us, so we always appreciate any help in this area.

I'm not 100% sure why that function always assumes there is a cmd shim to be honest. @MarshallOfSound  might have some sort of an idea. These functions were based on the implementations in `npm` and `lerna` if I'm not mistaken, so I'm a little bit cautious to change away from how they did it.

That being said, it definitely looks strange to me. I'm not sure on the *best* way to do this, but in the mean time, could you change the `var`s to `let`s and check the error code coming back from `readCmdShim`?

```
      if (err.code !== 'ENOTASHIM') {
        throw err;
      }
```

CC: @jamiebuilds as well, you'd probably have the most context as to why we always assume there's a cmd shim?

ajaymathur:
@MarshallOfSound Reproducible example repository shared in the issue:

https://github.com/kireerik/refo

from this [comment](https://github.com/boltpkg/bolt/issues/207#issuecomment-449865671)

Bloodb0ne:
What happens in really is when you have reference a standard package you get a source with a directory that's your normal  `/node_modules/.bin`  but otherwise it has a dir for example `/build/releases/bin/` where my .js file resides. My guess is that standard packages when they get installed the shim is auto created and that's not true for internal packages in a workspace.
I understand why you would be hesitant to merge when you use an existing implementation from npm to write that code.

Bloodb0ne:
Seems like the shim creation is a big issue in a other repos too https://github.com/yarnpkg/yarn/pull/6959

PR #224, opened 2019-04-22
Add new line to the end of file, preserve line endings style

pastelsky:
This PR fixes a couple of issues with bolt - 

- Bolt is opinionated about the line endings style – it always uses `\n` to denote line endings. This PR enables bolt to automatically detect line endings style (`\n` vs `\r\n`) using `detect-newline` and preserve the style when updating / writing to a package.json file

- Whenever bolt writes a package.json, it now adds a newline character to the last line, in accordance with the [POSIX definition of a line](https://stackoverflow.com/a/729795/3210476).

Note that these issues have already been fixed by `npm` ([here](https://github.com/npm/npm/pull/19904)) and `yarn` ([here](https://github.com/yarnpkg/yarn/commit/05bf97728cbe92f034f7d4596830f481a65fe8fa)), and this just brings bolt's implementation on par.

## Motivation
When using prettier or an opinionated editor alongside bolt, there is a constant push and pull between these tools that leads to ghost diffs like these –

image

### DIFF
--- a/src/Config.js
+++ b/src/Config.js
@@ -72,6 +72,7 @@ export default class Config {
   fileContents: string;
   json: JSONValue;
   indent: string;
+  newline: string;
   invalid: boolean;
 
   constructor(filePath: string, fileContents: string) {
--- a/src/Config.js
+++ b/src/Config.js
@@ -1,4 +1,5 @@
 // @flow
+import os from 'os';
 import pkgUp from 'pkg-up';
 import detectIndent from 'detect-indent';
 import detectNewline from 'detect-newline';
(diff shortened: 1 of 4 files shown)


src/Config.js
```diff
@@ -71,13 +72,15 @@ export default class Config {
...
   indent: string;
+  newline: string;
   invalid: boolean;
 
   constructor(filePath: string, fileContents: string) {
     this.filePath = filePath;
     this.fileContents = fileContents;
     try {
       this.indent = detectIndent(fileContents).indent || '  ';
+      this.newline = detectNewline(fileContents) || '\n';
```
ajaymathur: Should we use `os.EOL`? ( It is a question because I am sure if that has any advantage over having `'\n'` )

ajaymathur:
Hey @pastelsky , the change looks good. I have just added a comment for clarity :)

src/Config.js
pastelsky: `os.EOL` would be a dynamic value. I think the idea here is to keep the line-ending style in `package.json` consistent even if folks working on different OSes are making changes to this file.

pastelsky:
Updated the PR with `os.EOL`.

PR #227, opened 2019-05-06
add feedback on otp failure

Noviny:
I was looking at modifying changesets to prompt for OTP on this kind of failure, but wasn't getting enough information back from bolt. This should allow the desired change in changesets.


lukebatchelor:
LGTM, but the tests seem to fail in CI?

lukebatchelor (changes requested):
Code looks good, just need those tests passing

8012145393085:
Hello world

PR #229, opened 2019-05-13
Remove quotes around `yarn upgrade` args

### LINKED ISSUE #228: `bolt upgrade` should not wrap versioned packages with quotes
| Version             | at least v0.22.4 through v0.23.2
| Type                | Issue
| node                | v10.6.0, v12.2.0
| Operating System    | macOS Mojave 10.14.4
| Short Description   | wrapping versioned packages with `"` causes yarn to fail parsing package name correctly under certain conditions
| Detailed description| too long, I suppose, see below the table

Now, this one is truly weird, but easily reproducible:
1. Preconditions: you've got `node`, `bolt`, and `yarn`
2. ```sh
    mkdir test && cd test
    ```
3. ```sh
    yarn init -y
    ```
4. ```sh
    yarn add lodash@^3
    ```
   (could actually be any package)
5. ```sh
    node -e "require('child_process').execSync('bolt upgrade lodash@^4', { stdio: 'ignore' })"
    ```
    this will fail, which is **the bug**
6. ```sh
   cat yarn-error.log | grep Error
    ```
    >   Error: https://registry.yarnpkg.com/%22lodash: Not found

Now, there's not a lot going on here, but here's what else I was able to find:
1. `{ stdio: 'ignore' }` plays a role, the same thing will happen with `{ stdio: 'pipe' }`, but it actually **_succeeds_** with `{ stdio: 'inherit' }` 🤯
2. [This line](https://github.com/boltpkg/bolt/blob/master/src/utils/yarn.js#L86)seems to be the culprit - there's no need to quote-wrap arguments passed down to `spawn`
3. On yarn's side, [this function](https://github.com/yarnpkg/yarn/blob/eb2b565bb9b948e87b11119482ebc184a9d66141/src/util/normalize-pattern.js) is the blowup point - when `normalizePattern('"lodash@^4"')` is invoked, it produces the following: `{ name: '"lodash', range: '^4"', hasVersion: true }`

I'm very uncertain as to why `stdio: 'inherit'` succeeds, but it looks like quote-wrapping the args is redundant and can be safely removed.

DethAriel:
Fixes #228

### DIFF
--- a/src/utils/yarn.js
+++ b/src/utils/yarn.js
@@ -83,7 +83,7 @@ export async function upgrade(
   if (dependencies.length) {
     dependencies.forEach(dep => {
       if (dep.version) {
-        spawnArgs.push(`"${dep.name}@${dep.version}"`);
+        spawnArgs.push(`${dep.name}@${dep.version}`);
       } else {
         spawnArgs.push(dep.name);
       }


lukebatchelor:
LGTM, cheers!

lukebatchelor:
Released in bolt@0.23.3

PR #231, opened 2019-05-19
Don't read packages on each iteration

goodmind:
- Check for `package.json` file directly instead of looking for `package.json` in every folder
- Add `prepare` lifecycle because `prepublish` doesn't work with `git` dependency
- Don't read packages on each iteration


jamiebuilds:
Why was this closed?

goodmind:
Because I added irrevelant commits to this PR :\

goodmind:
@jamiebuilds this was related to this https://github.com/flow-typed/flow-typed/issues/3083

PR #232, opened 2019-05-22
Don't read packages on each iteration

goodmind:


goodmind:
There's also bug that it reads `node_modules` tho

PR #234, opened 2019-06-07
Fix fix for yarn setting an npm_config_registry

jamiebuilds:
Fixes https://github.com/boltpkg/bolt/pull/190#issuecomment-499965466

The problem with the original fix is that it makes the same mistake Yarn did by overwriting all registry configs. So I changed this to do the minimal amount of work to revert Yarn's incorrect behavior.

Note: Setting an env variable to `undefined` just completely eliminates it, which will cause npm to use its default behavior. Only when `npm_config_registry` has been set to something other than `registry.yarnpkg.com` should we respect it.

### DIFF
--- a/src/utils/__tests__/npm.test.js
+++ b/src/utils/__tests__/npm.test.js
@@ -50,7 +50,7 @@ describe('npm', () => {
         containDeep({
           cwd,
           env: {
-            npm_config_registry: 'https://registry.npmjs.org/'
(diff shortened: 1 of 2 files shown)


lukebatchelor:
Ah Cheers, theree definitely was an issue to come back and fix that.

goodmind:
Can we have `--registry` option?

PR #235, opened 2019-06-25
Updated to skip workspaces that are missing a package.json

kwelch:
Instead of sending a message about an empty workspace, skip it entirely. 

This avoids having to remove the directory and re-run the command. 

I added a test that covers this case as well.


kwelch:
This is still getting a single failure, but it is only on node v4. Honestly, I am not sure what I did that would have caused that.

src/Config.js
```diff
@@ -280,9 +280,8 @@ export default class Config {
     if (typeof bin === 'string') return bin;
     return toObjectOfStrings(
       bin,
-      `package.json#bin must be an object of strings or a string. See "${
-        this.filePath
-      }"`
+      `package.json#bin must be an object of strings or a string. See "${this
+        .filePath}"`
```
jamiebuilds: Is your editor not using the same version of Prettier as Bolt?

src/__tests__/Config.test.js
```diff
@@ -81,7 +81,7 @@ describe('writeConfigFile', () => {
     let json = { name: 'wat', version: '0.0.0' };
     let fileContents =
       JSON.stringify(json, null, 2).replace(/\n/g, '\r\n') + '\r\n';
-    let config = await Config.init(filePath);
+    let config = (await Config.init(filePath)) || {};
```
jamiebuilds: Why `|| {}`?

src/Config.js
kwelch: Possibly, let me check.

src/Config.js
```diff
@@ -98,13 +100,14 @@ export default class Config {
...
-  static async init(filePath: string): Promise<Config> {
+  static async init(filePath: string): Promise<?Config> {
     let fileContents;
     try {
       fileContents = await fs.readFile(filePath);
     } catch (e) {
       if (e.code === 'ENOENT') {
-        logger.error(messages.cannotInitConfigMissingPkgJSON(filePath));
+        // swallow error on missing config
+        return;
```
jamiebuilds: I don't think you should return here, I think you should throw an error and catch that error elsewhere

src/__tests__/Config.test.js
kwelch: Since it is returning a maybe it needs a default.

src/Config.js
kwelch: oh interesting, that changes the whole apporach since it is no longer going to return a maybe. 🤔

src/__tests__/Config.test.js
kwelch: moved to throwing an error instead of a maybe and this got better.

src/Config.js
kwelch: Throw bolt error now, then catching in getPackages. Tests all pass, but I am still slightly uneasy about this mainly because I don't understand the whole flow.

Issue #237, opened 2019-07-23
Feature request: Support yarn `--frozen-lockfile`


Offirmo:
| Title               | Description
| --------------------|------------
| Version             | 0.23.5
| Type                | feature Request
| node                | 10
| Short Description   | Support yarn `--frozen-lockfile`
| Detailed description| 

In the same way that `--pure-lockfile` is supported, we should have a way to support `--frozen-lockfile`, especially useful for CI.

I can't do the PR because the best way to change it is to turn `pureLockfile: boolean` into a more powerful `lockfileMode: 'default' | 'pure' | 'frozen'` but that's above my flow level :-(

lukebatchelor:
Hmm.. I'd think we'd just add it as a separate flag right? We pass it directly onto yarn anyway, so I don't think we'd deviate from their API?

I'd be happy to accept it if you like.

Offirmo:
Actually it wasn't that hard: https://github.com/boltpkg/bolt/pull/238

lukebatchelor:
Closed in #238

lukebatchelor:
Released in 0.23.6

PR #238, opened 2019-07-23
feat: allow passing --frozen-lockfile to yarn

Offirmo:
#237

refactored internally as an enum to be a better domain language than a set of booleans. Also those options are competing so we should have only one.

No regression from before, `--pure-lockfile` still works as expected.

### DIFF
--- a/src/commands/__tests__/install.test.js
+++ b/src/commands/__tests__/install.test.js
@@ -44,7 +44,7 @@ describe('install', () => {
       })
     );
     expect(yarn.install).toHaveBeenCalledTimes(1);
-    expect(yarn.install).toHaveBeenCalledWith(cwd, false);
+    expect(yarn.install).toHaveBeenCalledWith(cwd, 'default');
   });
 
   test('should still run yarn install at the root when called from ws', async () => {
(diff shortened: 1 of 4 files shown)


Offirmo:
node 4 is broken but I don't believe it's my fault

lukebatchelor:
Not sure I agree with this though.
I don't see any reason to stray from the existing API? This just makes it harder for consumers right?
Why not just take both (same as yarn) and pass them to yarn, so that yarn behaves exactly as it would if it were run directly?

Offirmo:
@lukebatchelor It seems there is a misunderstanding. I din't change the external API at all, it's still `--pure-lockfile` or `--frozen-lockfile`. The enum is just the **internal** representation of those flags while we are passing them around.

lukebatchelor:
Ha, I'm an idiot. Apologies.

lukebatchelor:
LGTM, node 4 looks broken on master too.

Cheers, I'll release it when I get to my work computer.

Offirmo:
@lukebatchelor I'm the idiot. I re-read your comment and it makes sense as well. I'm introducing a constraint (-frozen and -pure can't coexist) which is not present in the underlying tool (yarn).

You suggestion to just pass the options with no interference makes total sense as well.

lukebatchelor:
Released in 0.23.6

PR #198, opened 2018-09-28
Pass bolt_config_user_agent to env when running install

randing89:
`npm_config_user_agent` seems being overridden by `yarn`.
This PR exposes bolt version via environment variable.

PR #240, opened 2019-07-26
Add bolt_config_user_agent env var to `yarn add`

Blasz:
Follows on from #198 which added this to the install command. This
should also be added to at least the add command as preinstall yarn
scripts are ran for both install and add.

### DIFF
--- a/src/utils/__tests__/yarn.test.js
+++ b/src/utils/__tests__/yarn.test.js
@@ -35,13 +35,13 @@ function assertSpawnCalls(expectedProcess, expectedArgs, expectedCwd) {
 let yarnUserAgent = 'yarn/7.7.7 npm/? node/v8.9.4 darwin x64';
 
 describe('utils/yarn', () => {
+  beforeEach(() => {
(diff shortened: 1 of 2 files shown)


lukebatchelor:
Apologies for letting this one slip through @Blasz .

LGTM. Only thought is, is there anyway we could make this just always be set whenever we run anything through spawn?  Or am I just missing some context here and this particular case is special?

Give me a shout and I'll merge.

Blasz:
@lukebatchelor Yeah we could probably do that, nothing special about this particular case that I'm aware of. I don't have a ton of context surrounding this either.

I'll implement it sometime soon.

Blasz:
@lukebatchelor I've updated the implementation to add the user agent vars to all yarn commands. We spawn some other non-yarn commands so I didn't want to add it to all spawn calls, I think yarn is a decent trade-off.
Let me know what you think.

lukebatchelor:
Yup, completely agree.

lukebatchelor:
Released in 0.24.2 🎉

PR #241, opened 2019-08-16
Fix windows compatibility Issue #207

bholloway:
Alternative fix for #207 instead of existing PR #215.

**src/utils/fs.js**

In the `readLink` method there is a comment [here](https://github.com/boltpkg/bolt/blob/master/src/utils/fs.js#L135-L136) stating code was copied from `npm` cli [here]().

The copied code doesn't have a fallback when the file is _not_ a symlink.

However when we look further in `npm` cli. The copied code is used in such a way [here](https://github.com/npm/npm/blob/d081cc6c8d73f2aa698aab36605377c95e916224/lib/utils/gently-rm.js#L273) as there _is_ a fallback similar to what @Bloodb0ne has proposed in #215.

In #215 the `try/catch` is duplicated in `cmdShim` whereas here I simply use the already safe `readLink` method.

~In this PR I also removed the unused `path-is-inside` dependency.~

**src/utils/symlinkPackageDependencies.js**

In my testing I also received`... is not a symlink` error in `symlinkPackageDependencies` for workspace packages that exposed bin scripts. In lerna these shims are not symlinks [here](https://github.com/lerna/lerna/blob/8ca18bedecf4f141c6242a099086e84b2ced72de/utils/resolve-symlink/resolve-symlink.js#L49) and considered to be packages that were erroneously installed. They are removed [here](lerna/lerna/blob/8ca18bedecf4f141c6242a099086e84b2ced72de/utils/symlink-dependencies/symlink-dependencies.js#L64) before relinking.

In `bolt` I suspect it is correct to read the shim as if it is a symlink and continue, similar to what is done in `cmdShim` (as discussed above). From what I can see `bolt` then tries to recreate the shim [here](https://github.com/boltpkg/bolt/blob/master/src/utils/fs.js#L89) which is presumably idempotent.

### DIFF
--- a/src/utils/fs.js
+++ b/src/utils/fs.js
@@ -57,10 +57,12 @@ function stripExtension(filePath: string) {
 }
 
 async function cmdShim(src: string, dest: string) {
-  const currentShimTarget = path.resolve(
-    path.dirname(src),
-    await readCmdShim(src)
-  );
+  // If not a symlink we default to the actual src file
+  // https://github.com/npm/npm/blob/d081cc6c8d73f2aa698aab36605377c95e916224/lib/utils/gently-rm.js#L273
+  let relativeShimTarget = await readlink(src);
+  let currentShimTarget = relativeShimTarget
+    ? path.resolve(path.dirname(src), relativeShimTarget)
+    : src;
   await promisify(cb => _cmdShim(currentShimTarget, stripExtension(dest), cb));
 }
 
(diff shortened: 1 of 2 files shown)


bholloway:
Ping @lukebatchelor

bholloway:
The CI is failing Node 4. Seems to be a [known issue](https://github.com/yarnpkg/yarn/issues/6619) with yarn@latest.

lukebatchelor:
LGTM! Thanks heaps @bholloway. Let me run this past Jamie quickly and we'll get this merged and released.

lukebatchelor:
Ping @jamiebuilds . Looks G to me. You happy to merge?

lukebatchelor:
Thanks heaps @bholloway 

Released in 0.24.1

PR #242, opened 2019-08-27
fix #56 for workspaces --dev flag swallows first dependency

### LINKED ISSUE #56: bolt add --dev pkgName does not seem to work

ForbesLindesay:
This is exactly the same as #67 for the `bolt w X add -D dev-dependency` command.


lukebatchelor:
Oh, awesome work. I actually have seen this bug in the wild and assumed I typo'd somewhere.

Cheers! Will merge now and update here when its released (will batch with a couple of PRs this week)

PR #244, opened 2019-09-02
Allow exclusion of certain dependency types from dependency graph

Blasz:
This adds the `--excludeFromGraph` cli flag to `bolt workspaces exec` and other commands that accept `spawnOpts` to optionally exclude certain dependency types from being included in the dependency graph that is used to execute commands in a topological order.

As part of this change, the `getDependencyGraph` and `getDependentsGraph` functions now also accept an optional `excludedTypes` key in their `opts` argument to exclude dependency types as well.

Our primary use case for adding this flag is to filter out devDependencies when executing our build as we only care about dependencies/peerDependencies. When we include devDependencies we get unavoidable cycles which then result in commands being executed out of order.


lukebatchelor:
Hey @jamiebuilds Just wanted to get your thoughts on this one before merging it.
We're not _super_ happy with it, it feels a bit "tacked on", so more than open to other approaches. The idea itself though is pretty important imo. It's causing a bunch of actual errors on our end due to to cycles that don't _reeeaaallly_ need to be considered.

Happy to iterate though.

lukebatchelor:
Sweet, think I'll merge this now. Thanks @Blasz !

lukebatchelor:
Released in 0.24.3

PR #248, opened 2019-09-23
Update toDependency util to handle versions containing @ symbols

### LINKED ISSUE #239: Local paths containing '@' aren't handled properly
| Version             | v0.23.3
| Type                | Issue
| node                | v10.9.0

Ran into a gnarly bug where running an upgrade command with a local path that includes an '@' chops off the path afterwards.
E.g. `bolt upgrade @atlaskit/button@file:.yalc/@atlaskit/button` results in the version being parsed as `file:.yalc/` due to https://github.com/boltpkg/bolt/blob/master/src/utils/options.js#L68 splitting on '@'.

I'm running this problem after:
1. Linking an external dependency via yalc - which updates package.json to a local path containing the package name
2. Running `bolt` afterwards to reinstall new dependencies after pulling master
3. Bolt workspace validation complains after noticing version discrepancies between the local path in the root and each workspace containing the version that was installed previously
4. Running the upgrade command above to enforce them to all be the same local path

It's a bit convoluted since the better way to fix the bolt validation errors is to just revert the yalc changes, run bolt and then re-apply them.

Blasz:
E.g. local file path installs could have @ symbols, especially if the
local path contains the fully scoped package name.

Fixes #239.

### DIFF
--- a/src/utils/__tests__/options.test.js
+++ b/src/utils/__tests__/options.test.js
@@ -48,6 +48,18 @@ describe('options', () => {
     expect(toDependency(nameWithVersion)).toEqual({ name: nameWithVersion });
     nameWithVersion = `@packageName@${version}`;
     expect(toDependency(nameWithVersion)).toEqual({ name, version });
+    // Version with @ symbol
+    expect(toDependency(`packageName@file:local/@my-packages`)).toEqual({
+      name: 'packageName',
+      version: 'file:local/@my-packages'
+    });
+    // Scoped package with @ symbol version
+    expect(
+      toDependency(`@scope/packageName@file:local/@scope/packageName`)
+    ).toEqual({
+      name: '@scope/packageName',
+      version: 'file:local/@scope/packageName'
+    });
   });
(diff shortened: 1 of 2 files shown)


src/utils/__tests__/options.test.js
```diff
@@ -48,6 +48,18 @@ describe('options', () => {
     expect(toDependency(nameWithVersion)).toEqual({ name: nameWithVersion });
     nameWithVersion = `@packageName@${version}`;
     expect(toDependency(nameWithVersion)).toEqual({ name, version });
+    // Version with @ symbol
+    expect(toDependency(`packageName@file:local/@my-packages`)).toEqual({
```
lukebatchelor: I found this example slightly confusing, but as explained irl, its looking at an unscoped package name with a path that contains an '@'. Super unlikely to ever happen, but an important code path to double check that the logic is correct.

lukebatchelor:
Released in 0.24.2 🎉

PR #250, opened 2019-09-23
Upgrade prettier to latest

Blasz:
The current version of prettier (1.7.0) had a bug that was breaking the pre-commit hook so I've upgraded it to the latest.


lukebatchelor:
Released in 0.24.2 🎉

PR #251, opened 2019-10-01
Update workspace versions when running bolt add on an existing dependency in the project root

Blasz:
We updated the `bolt add` command in the last version to re-run `yarn add` when run from the project root to allow installing new versions of an existing dependency without needing to run `bolt upgrade` which behaves differently (it recursively upgrades transitive deps).

As part of this work, we forgot to also update the versions of the dependency being added in each workspace in the repo. Resulting in version mismatch errors on subsequent runs of bolt until workspace versions were manually updated.


lukebatchelor:
Released in 0.24.3

PR #252, opened 2019-10-02
Implement help command and modify --help behaviour

Blasz:
--help now only displays help information if ran with no commands, i.e.
`bolt --help`.
If a command is passed, e.g. bolt <command> --help`. the `--help` flag will be
passed through to that command and not display help text.
This fixes running npm scripts that have implemented their own flag,
e.g. `yarn run my-script-that-has-a-help-flag --help`

In the future we will implement help information for subcommands that
may handle `bolt <command> --help` as well as `bolt help <subcommand>`.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -412,7 +412,7 @@ function runCommandFromCli(args: options.Args, flags: options.Flags) {
 export default async function cli(argv: Array<string>, exit: boolean = false) {
   let start = Date.now();
 
-  let { pkg, input, flags } = meow({
+  let { pkg, input, flags } = meow('', {
     argv,
     flags: {
       '--': true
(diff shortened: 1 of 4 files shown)


src/cli.js
```diff
@@ -408,12 +412,12 @@ function runCommandFromCli(args: options.Args, flags: options.Flags) {
 export default async function cli(argv: Array<string>, exit: boolean = false) {
   let start = Date.now();
 
-  let { pkg, input, flags } = meow({
+  let { pkg, input, flags } = meow('', {
```
lukebatchelor: Whats the string here for @Blasz ? 🤔 💭

Blasz: The `meow` flow typedefs don't support the variant without the helptext as the first arg so I just gave it an empty string.
This is what `meow` itself does when passing the options as the first argument anyway. 
The reason why we don't pass the full help text is we're not relying on meow to display help text.

lukebatchelor:
Released in 0.24.3

PR #254, opened 2019-10-02
fix(Yarn): replaced relative path to bin with npx and first argument

Stradivario:
# [Trello Card](https://trello.com/c/7jX54Hu9/6777-failing-migration-hard-to-debug)

# Description
This issue present when we use Codeship Build process. Codeship environment has different way of linking, mapping, caching installed packages so it is not a good option to think that we will have yarn installed and linked for us. This pull request introduces simple fix using npx command.

## Type of change

- [x] Non Breaking change

# Checklist:

- [x] Replace relative path of `yarn` with `npx` command
- [x] Put `yarn` command inside first argument of execution inside child process

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,7753 @@
+{
+  "name": "bolt",
+  "version": "0.24.2",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "abab": {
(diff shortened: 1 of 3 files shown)


Stradivario:
@lukebatchelor  @Blasz

PR #256, opened 2019-10-23
Fix --no-bail not throwing when running in default orderMode

### LINKED ISSUE #255: --no-bail flag does not exit with errors when using default order mode
The --no-bail flag does not throw when executing tasks in the default parallelGraph mode.

To reproduce: `bolt ws exec --no-bail -- false`

Whereas `bolt ws exec --no-bail --parallel -- false` works as intended

Blasz:
Fixes #255.

I've also improved the error message to report the number of tasks that failed and any error messages that were in the error (which is most likely going to be none for the cli version since it spawns processes).

### DIFF
--- a/src/Project.js
+++ b/src/Project.js
@@ -268,16 +268,21 @@ export default class Project {
       graph.set(pkgName, pkgInfo.dependencies);
     }
 
-    let { safe, values } = await taskGraphRunner({
-      graph,
-      force: true,
-      task: async pkgName => {
-        let pkg = this.getPackageByName(packages, pkgName);
-        if (pkg) {
-          return task(pkg);
+    let {
+      safe,
+      values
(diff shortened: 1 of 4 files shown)


src/Project.js
```diff
@@ -197,12 +198,18 @@ export default class Project {
...
+    const taskFailures: Array<PromiseFailure> = (results.filter(
+      r => r && r.status === 'error'
+    ): any);
+    if (taskFailures.length > 0) {
+      const failuresWithMsg = taskFailures
+        .map(r => r.error && r.error.message)
+        .filter(Boolean)
+        .join('\n');
+      throw new BoltError(
+        `${taskFailures.length} tasks failed.\n${failuresWithMsg}`
```
lukebatchelor: Should this live in the messages file?

lukebatchelor:
Sweet, LGTM

PR #283, opened 2020-10-19
Fix workspace versions not resolving when installed via dist tag

### LINKED ISSUE #274: Bolt add fails when using dist-tag versions if dependency already exists in workspaces
We changed bolt add recently to work if the dependency already existed in the repo, similar to yarn.

When using dist-tag versions, e.g. `bolt add foo@latest`, it will update workspace versions of that dependency to 'latest' rather than the resolved version. The original dependency version in the root package.json is correct.

I've confirmed that `bolt upgrade` works correctly so it's just an issue with `bolt add`.

Blasz:
Fixes #274.

I've ran into this a bunch of times but have never got around to fixing it.

### DIFF
--- a/src/utils/__tests__/addDependenciesToPackages.test.js
+++ b/src/utils/__tests__/addDependenciesToPackages.test.js
@@ -16,13 +16,23 @@ jest.mock('../logger');
 
 const unsafeYarn: any & typeof yarn = yarn;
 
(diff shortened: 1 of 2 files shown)


lukebatchelor:
Nice one 🎉

lukebatchelor (approved):

lukebatchelor:
Released in `0.24.7`

PR #286, opened 2020-12-08
updates for faster symlinking

jonathancreamer-eb:
Rather than re-fetch the Project packages, and graph during the for of loop from the install command, simply passing them in saves a ton of time.

Converts the for of loop for symlinking into a Promise.all.

Don't re-create symlinks.

In our project we have ~360 internal packages, and ~370 external.

Before...

After..

### DIFF
--- a/src/utils/symlinkPackageDependencies.js
+++ b/src/utils/symlinkPackageDependencies.js
@@ -14,15 +14,25 @@ import * as yarn from './yarn';
 export default async function symlinkPackageDependencies(
   project: Project,
   pkg: Package,
-  dependencies: Array<string>
+  dependencies: Array<string>,
+  pkgs: Array<Packages>,
+  dependencyGraph,
 ) {
-  let projectDeps = project.pkg.getAllDependencies();
-  let pkgDependencies = project.pkg.getAllDependencies();
-  let packages = await project.getPackages();
-  let {
-    graph: dependencyGraph,
-    valid: dependencyGraphValid
-  } = await project.getDependencyGraph(packages);
+
+  let packages = pkgs || await project.getPackages();
+  
+  let valid;
+  let dependencyGraphValid = true;
+
+  if (!dependencyGraph) {
+    let {
+      graph,
+      valid,
+    } = await project.getDependencyGraph(packages);
+    dependencyGraph = graph;
+    dependencyGraphValid = valid;
+  }
+  
   let pkgName = pkg.config.getName();
   // get all the dependencies that are internal workspaces in this project
   let internalDeps = (dependencyGraph.get(pkgName) || {}).dependencies || [];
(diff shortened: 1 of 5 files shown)


jonathancreamer-eb:
@lukebatchelor if you get a chance soon, I'd love for you to take a look at this. It was a pretty huge difference for us in our repo on my initial testing. Let me know what you think!

aswensen-eb:
This seems to improve package performance by 92%. 🎉

src/utils/symlinkPackageDependencies.js
```diff
@@ -14,30 +14,41 @@ import * as yarn from './yarn';
...
 ) {
-  let projectDeps = project.pkg.getAllDependencies();
-  let pkgDependencies = project.pkg.getAllDependencies();
-  let packages = await project.getPackages();
-  let {
-    graph: dependencyGraph,
-    valid: dependencyGraphValid
-  } = await project.getDependencyGraph(packages);
+
+  let packages = pkgs || await project.getPackages();
```
jamiebuilds, line 22: Whats going on with this `||`... is the type of `pkgs` wrong?

jonathancreamer-eb: No, it supports either passing the `pkgs` in or fetching them. That was mostly a backwards compatible thing, in case this is called from a different flow. I was primarily fixing the install step.

jamiebuilds: Do you want to validate if it is called in any other flows? I don't think it's hard to require people to call `getPackages()` first

jonathancreamer-eb: The only other place I can tell it's called is here...

https://github.com/boltpkg/bolt/blame/master/src/utils/addDependenciesToPackages.js#L100

When I passed the packages and graph in there though, it failed a bunch of tests in the addDependenciesToPackage tests.

jonathancreamer-eb: I'll see if I can dig into why it's failing. Has something to do with the symlinkSpy.

jonathancreamer-eb: I'm now realizing that the only reason the symlinkDependencies function needed the packages was to pass into the graph. So, updating now to only pass the graph into the symlinkDependencies.

jonathancreamer-eb:
@jamiebuilds updated! c0131ab63200136cb1b87a20ab4a9c8c44d4182a

lukebatchelor:
Thanks heaps @jonathancreamer-eb , super appreciate the investigation. 
PR LGTM, I'll wait and see if Jamie has any other feedback, and get this released soon.

lukebatchelor (approved):

jonathancreamer-eb:
Thanks! Hopefully this helps all of us. :)

jonathancreamer-eb:
Oh hai 2021. Shall we start it off with faster `bolt` or no @jamiebuilds? ;)

